PR #15, opened 2015-01-02
Fix that prevents variables from being defined and assigned in a single ...

gitlarryf:
...inline statement.

### DIFF
--- a/src/parser.cpp
+++ b/src/parser.cpp
@@ -1323,7 +1323,7 @@ const Statement *Parser::parseReturnStatement(Scope *scope, int line)
 const Statement *Parser::parseVarStatement(Scope *scope, int line)
 {
     ++i;
-    std::vector<const ReferenceExpression *> varrefs;
+    std::vector<Variable *> variables;
     const VariableInfo vars = parseVariableDeclaration(scope);
     for (auto name: vars.first) {
         Variable *v;
@@ -1332,15 +1332,24 @@ const Statement *Parser::parseVarStatement(Scope *scope, int line)
         } else {
             v = new LocalVariable(name, vars.second, scope, false);
         }
-        scope->addName(name, v);
-        varrefs.push_back(new VariableExpression(v));
+        variables.push_back(v);
     }
+    const Statement *r = nullptr;
     if (tokens[i].type == ASSIGN) {
         ++i;
+        std::vector<const ReferenceExpression *> refs;
         const Expression *expr = parseExpression(scope);
-        return new AssignmentStatement(line, varrefs, expr);
+        for (auto v: variables) {
+            scope->addName(v->name, v, true);
+            refs.push_back(new VariableExpression(v));
+        }
+        r =  new AssignmentStatement(line, refs, expr);
+    } else {
+        for (auto v: variables) {
+            scope->addName(v->name, v);
+        }
     }
-    return nullptr;
+    return r;
 }
 
 const Statement *Parser::parseWhileStatement(Scope *scope, int line)
--- a/t/inline-init4.neon
+++ b/t/inline-init4.neon
@@ -1,0 +1,5 @@
+% NOTE the following test case does not return any values, 
+% as this specific test is testing that the variable
+% is referenced once created.
+
+VAR a: Number := 5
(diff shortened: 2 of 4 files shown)


ghewgill:
This doesn't fix the problem of running the program

    VAR a: Number := 5

src/parser.cpp
```diff
@@ -1332,14 +1333,21 @@ const Statement *Parser::parseVarStatement(Scope *scope, int line)
         } else {
             v = new LocalVariable(name, vars.second, scope, false);
         }
-        scope->addName(name, v);
+        variables.insert(std::map<std::string, Variable *>::value_type(name, v));
+
```
ghewgill: Remove blank line.

src/parser.cpp
```diff
@@ -1332,14 +1333,21 @@ const Statement *Parser::parseVarStatement(Scope *scope, int line)
         } else {
             v = new LocalVariable(name, vars.second, scope, false);
         }
-        scope->addName(name, v);
+        variables.insert(std::map<std::string, Variable *>::value_type(name, v));
```
ghewgill: Rather than using `insert` with a pair, just do: `variables[name] = v;`

src/parser.cpp
```diff
@@ -1332,14 +1333,21 @@ const Statement *Parser::parseVarStatement(Scope *scope, int line)
...
+        
         varrefs.push_back(new VariableExpression(v));
```
ghewgill: You're setting up `varrefs` here but only use it in the case where an assignment is happening. Suggest just creating a list of `Variable *`, as `variables` that you already have. Then, in the case where you need to create an `AssignmentStatement`, create `varrefs` at that point.

src/parser.cpp
```diff
@@ -1332,14 +1333,21 @@ const Statement *Parser::parseVarStatement(Scope *scope, int line)
...
         const Expression *expr = parseExpression(scope);
+        for (auto v: variables) {
+            scope->addName(v.first, v.second, true);
+        }
         return new AssignmentStatement(line, varrefs, expr);
     }
+    for (auto v: variables) {
+        scope->addName(v.first, v.second);
+    }
     return nullptr;
```
ghewgill: I'm a bit uncomfortable with this logic for managing the return value. I might suggest something like:

    const Statement *r = nullptr;
    if (tokens[i].type == ASSIGN) {
        // assign stuff, set r to an AssignmentStatement
    } else {
        // normal addName without setting the referenced bit
    }
    return r;

t/inline-init4.neon
```diff
@@ -0,0 +1 @@
+VAR a: Number := 5
```
ghewgill: Add a comment here saying that there is no output tested because you're specifically testing a bug that happened when the variable was *not* referenced later.

src/parser.cpp
```diff
@@ -1323,7 +1323,7 @@ const Statement *Parser::parseReturnStatement(Scope *scope, int line)
 const Statement *Parser::parseVarStatement(Scope *scope, int line)
 {
     ++i;
-    std::vector<const ReferenceExpression *> varrefs;
+    std::map<std::string, Variable *> variables;
```
ghewgill: I'd like to see this remain a vector. The reason is, the order of items in a map is not consistent. This could cause the same source file, when compiled more than once, to result in different bytecode (especially if built on a different machine). You can still access the name of the variable with `v->name` (for the `addName` calls below).

src/parser.cpp
```diff
@@ -1332,15 +1332,24 @@ const Statement *Parser::parseVarStatement(Scope *scope, int line)
         } else {
             v = new LocalVariable(name, vars.second, scope, false);
         }
-        scope->addName(name, v);
-        varrefs.push_back(new VariableExpression(v));
+        variables[name] = v;
```
ghewgill: With a vector, this will need to change to just `variables.push_back(v);`.

t/inline-init4.neon
```diff
@@ -0,0 +1,3 @@
+% NOTE the following test case does not return any values, as this specific test is testing that the variable is referenced once created.
```
ghewgill: A line break would be nice here :)

PR #16, opened 2015-01-06
Added standalone Neon bytecode disassembler

gitlarryf:


ghewgill:
Two general comments:

- Check spelling of "disassembler"
- Make the commit message shorter. No need to mention "Neon" in the commit message for example, since that's where the commit is. Suggest just "Add standalone disassembler".

src/neondis.cpp
```diff
@@ -0,0 +1,41 @@
...
+
+int main(int argc, char *argv[])
+{
+    if (argc < 2) {
+        fprintf(stderr, "Usage: %s file1.neonx file2.neonx...fileN.neonx\n", argv[0]);
+        exit(1);
+    }
+
+    int a = 1;
+    while (a < argc) {
```
ghewgill: No need for a loop here. It doesn't make sense to disassemble more than one file, especially when they are all written to stdout. Remember the unix mantra: "Do one thing, and do it well."

src/neondis.cpp
```diff
@@ -0,0 +1,41 @@
...
+        std::cout << "Disassembeling " << argv[a] << "...\n";
```
ghewgill: You don't need to print a status message if the thing you're about to do writes to stdout anyway. Remove this.

src/neondis.cpp
```diff
@@ -0,0 +1,41 @@
...
+
+        const std::string name = argv[a];
+
+        std::ifstream inf(name, std::ios::binary);
+        std::stringstream buf;
+        buf << inf.rdbuf();
+
+        // ToDo: Add Script header for better byte code detection.
```
ghewgill: Please use "TODO" comments, not "ToDo". Consistency.

src/neondis.cpp
```diff
@@ -0,0 +1,41 @@
...
+            fprintf(stderr, "%s is not a neonx file.\n", argv[a]);
+            ++a;
+            continue;
+        }
+
+        std::vector<unsigned char> bytecode;
+        std::string s = buf.str();
+        std::copy(s.begin(), s.end(), std::back_inserter(bytecode));
+
+        // ToDo: Detect Debug info stored in Bytecode, and pass pointer to it.
```
ghewgill: This comment is incorrect, debug info will not be store in the .neonx file. Just say something "TODO: Load debug info and pass a pointer to it." which doesn't imply any particular implementation.

src/neondis.cpp
```diff
@@ -0,0 +1,41 @@
...
+        disassemble(bytecode, std::cerr, NULL);
```
ghewgill: Since disassembly is the primary purpose of this program, the output should go to stdout, not stderr.

SConstruct
```diff
@@ -145,6 +145,16 @@ neonx = env.Program("bin/neonx", [
 ] + coverage_lib,
 )
 
+neondis = env.Program("bin/neondis", [
+    "src/bytecode.cpp",
+    "src/debuginfo.cpp",
+    "src/disassembler.cpp",
+    "src/number.cpp",
+    "src/util.cpp",
+    "src/neondis.cpp",
```
ghewgill: Alphabetical order for source files.

PR #23, opened 2015-01-20
Added Bytecode address offset on exception

gitlarryf:


src/exec.cpp
```diff
@@ -209,8 +211,8 @@ class Executor {
     Executor &operator=(const Executor &);
 };
 
-Executor::Executor(const Bytecode::bytecode &bytes)
-  : obj(bytes),
+Executor::Executor(const Bytecode::bytecode &bytes, const DebugInfo *debuginfo)
+  : obj(bytes), debug(debuginfo),
```
ghewgill: Put these on separate lines like the other instance variables

src/exec.cpp
```diff
@@ -979,7 +982,23 @@ void Executor::raise(const std::string &exception, const std::string &info)
         ip = callstack.top();
         callstack.pop();
     }
-    fprintf(stderr, "unhandled exception %s (%s)\n", exception.c_str(), info.c_str());
+    fprintf(stderr, "unhandled exception %s (%s) at address %lu\n", exception.c_str(), info.c_str(), eip);
+    
+    if (debug != nullptr) {
+        auto line = debug->line_numbers.find(eip);
+        if (line == debug->line_numbers.end()) {
```
ghewgill: The nesting here seems too deep. You should be able to do this with just a single `while` statement, and not an `if/for` nested block.

src/exec.cpp
```diff
@@ -979,7 +982,23 @@ void Executor::raise(const std::string &exception, const std::string &info)
...
+            for (;;) {
+                line = debug->line_numbers.find(--p);
+                if (line != debug->line_numbers.end()) {
+                    break;
+                }
+            }
+        }
+        fprintf(stderr, "%d | %s\n", line->second, debug->source_lines.at(line->second).c_str());
+    } else {
+        fprintf(stderr, "Neon binary was not built with additional debug information so no detailed info can be displayed.\n");
```
ghewgill: This message is way too long, also it's wrong. Just because no debug information is available *here*, doesn't mean the file was not built with debug information. Suggest simply `"No debug information available"` for now.

src/main.cpp
```diff
@@ -44,7 +45,10 @@ static void repl(int argc, char *argv[])
             exit(0);
         } else {
             try {
-                exec(compile(parse(tokenize(s)), nullptr), argc, argv);
+                auto ast = parse(tokenize(s));
```
ghewgill: You're going to have to rebase this against the current code before it can be merged.

src/main.cpp
```diff
@@ -95,6 +99,9 @@ int main(int argc, char *argv[])
     }
 
     std::vector<unsigned char> bytecode;
+    // TODO - Allow reading debug code from different source for in the event we are not
```
ghewgill: Can you make this comment shorter? I prefer to keep TODO comments to one line, so that a grep shows the whole message. Suggest `// TODO: Read debug info from file if needed.`.

src/neonx.cpp
```diff
@@ -27,7 +27,8 @@ int main(int argc, char *argv[])
     std::string s = buf.str();
     std::copy(s.begin(), s.end(), std::back_inserter(bytecode));
 
-    exec(bytecode, argc-1, argv+1);
+    // ToDo: Implement DebugInfo, for debugging compiled builds.
+    exec(bytecode, nullptr, argc-1, argv+1);
```
ghewgill: Should there also be a TODO comment in this file, just like in main.cpp?

src/exec.cpp
```diff
@@ -209,8 +211,9 @@ class Executor {
     Executor &operator=(const Executor &);
 };
 
-Executor::Executor(const Bytecode::bytecode &bytes)
-  : obj(bytes),
+Executor::Executor(const Bytecode::bytecode &bytes, const DebugInfo *debuginfo)
+  : obj(bytes),
```
ghewgill: This change adds a trailing space to the end of this line.

src/main.cpp
```diff
@@ -39,6 +39,7 @@ static void repl(int argc, char *argv[])
     for (;;) {
         std::cout << "> ";
         std::string s;
+        std::stringstream buf;
```
ghewgill: What is buf for here?

src/exec.cpp
```diff
@@ -979,7 +983,18 @@ void Executor::raise(const std::string &exception, const std::string &info)
...
         callstack.pop();
     }
-    fprintf(stderr, "unhandled exception %s (%s)\n", exception.c_str(), info.c_str());
+    fprintf(stderr, "unhandled exception %s (%s) at address %lu\n", exception.c_str(), info.c_str(), eip);
+    
+    if (debug != nullptr) {
+        auto line = debug->line_numbers.find(eip);
+        auto p = eip;
+        while (line == debug->line_numbers.end()) {
+            line = debug->line_numbers.find(--p);
```
ghewgill: I'm pretty sure you can further improve this loop. There are two calls to `.find()`, but there only needs to be one.

src/exec.cpp
```diff
@@ -979,7 +983,18 @@ void Executor::raise(const std::string &exception, const std::string &info)
...
+        }
+        fprintf(stderr, "%d | %s\n", line->second, debug->source_lines.at(line->second).c_str());
+    } else {
+        fprintf(stderr, "No debug information available.\n");
```
ghewgill: Add a TODO comment here for including the file name.

src/exec.h
```diff
@@ -3,6 +3,8 @@
 
 #include <vector>
 
-void exec(const std::vector<unsigned char> &obj, int argc, char *argv[]);
+#include "debuginfo.h"
```
ghewgill: Use a forward declaration here (`class DebugInfo;`) instead of including the whole header when it's not needed.

src/main.cpp
```diff
@@ -100,11 +104,13 @@ int main(int argc, char *argv[])
     }
 
     std::vector<unsigned char> bytecode;
+    // TODO - Allow reading debug information from file.
+    DebugInfo debug(buf.str());
 
     // Pretty hacky way of checking whether the input file is compiled or not.
     if (name[name.length()-1] != 'x') {
-
```
ghewgill: There seem to be a lot of changes to whitespace in this function that wasn't there in the last pull request. Don't make gratuitous changes to whitespace, as it obscures the real purpose of the commit.

src/exec.cpp
```diff
@@ -979,7 +983,19 @@ void Executor::raise(const std::string &exception, const std::string &info)
...
         callstack.pop();
     }
-    fprintf(stderr, "unhandled exception %s (%s)\n", exception.c_str(), info.c_str());
+    fprintf(stderr, "unhandled exception %s (%s) at address %lu\n", exception.c_str(), info.c_str(), eip);
+
+    if (debug != nullptr) {
+        auto p = eip;
+        auto line = debug->line_numbers.end();
+        while (line == debug->line_numbers.end()) {
+            line = debug->line_numbers.find(--p);
```
ghewgill: This loop will fail to find an entry in `line_numbers` that matches `eip` exactly, because `--p` will decrement `p` before calling `.find()`.

src/main.cpp
```diff
@@ -136,5 +145,8 @@ int main(int argc, char *argv[])
         std::copy(s.begin(), s.end(), std::back_inserter(bytecode));
     }
 
-    exec(bytecode, argc-a, argv+a);
+    exec(bytecode, debug, argc-a, argv+a);
+    if (debug != nullptr) {
+        delete debug;
```
ghewgill: Do not bother to `delete` this. The program is exiting anyway. See https://programmers.stackexchange.com/questions/140483/is-it-a-waste-of-time-to-free-resources-before-i-exit-a-process and http://blogs.msdn.com/b/oldnewthing/archive/2012/01/05/10253268.aspx

src/main.cpp
```diff
@@ -100,6 +104,8 @@ int main(int argc, char *argv[])
     }
 
     std::vector<unsigned char> bytecode;
+    // TODO - Allow reading debug information from file.
+    DebugInfo *debug = nullptr;
```
ghewgill: There's not actually any reason why this needs to be a pointer, is there? What was wrong with it being a local variable?

gitlarryf: Well, the DebugInfo has an explicit constructor, and no = operator, and no initialization method, so it really has to be created by an explicit call to the constructor, but we need to use it outside of the scope it's CREATED in, so, I figured the best way to do this is with a pointer...

src/exec.cpp
```diff
@@ -984,6 +987,7 @@ void Executor::raise(const std::string &exception, const std::string &info)
     globals[0].array()[0] = Cell(exception);
     globals[0].array()[1] = Cell(info);
     globals[0].array()[2] = Cell(number_from_uint32(static_cast<uint32_t>(ip)));
+    auto eip = ip;
```
ghewgill: Move this down a line (after the blank line) because it relates to the following code, not to the initialization of the exception data.

src/exec.cpp
```diff
@@ -998,7 +1002,19 @@ void Executor::raise(const std::string &exception, const std::string &info)
...
         callstack.pop();
     }
-    fprintf(stderr, "unhandled exception %s (%s)\n", exception.c_str(), info.c_str());
+    fprintf(stderr, "unhandled exception %s (%s) at address %lu\n", exception.c_str(), info.c_str(), eip);
+
+    if (debug != nullptr) {
+        auto p = eip;
+        auto line = debug->line_numbers.end();
+        while (line == debug->line_numbers.end()) {
+            line = debug->line_numbers.find(p--);
```
ghewgill: Don't use the inline decrement operator. I find this really hard to read and to think about whether it's correct or not.

PR #24, opened 2015-02-14
Added binary file reader RTL function and test.

gitlarryf:


ghewgill:
No longer required after introduction of `file.bytes` and the `Bytes` type.

PR #51, opened 2015-05-14
Fixed SCons build files to operate with older version of Python 2.6.6.

gitlarryf:


SConstruct
```diff
@@ -4,6 +4,16 @@ import tarfile
...
 
+def check_output(s, **kwargs):
+    p = subprocess.Popen(s, stdout=subprocess.PIPE, **kwargs)
+    (outdata, errdata) = p.communicate()
+    r = p.returncode
+    assert r is not None
+    if r != 0:
+        raise subprocess.CalledProcessError(r, s)
+    return outdata
+subprocess.check_output = check_output
```
ghewgill: This will *always* overwrite the `subprocess.check_output` function with the provided substitute. This should only take effect if the imported subprocess module does *not* already have a check_output function.

PR #60, opened 2015-06-25
Modified mkdir test to newer module standards.

gitlarryf:


gitlarryf:
I didn't want to move the permission denied, in case that exception was used for other things.  Like, it might not always be FILE SYSTEM error, so, I didn't want to limit it to just the file module, however I can certainly move it to there...

ghewgill:
Please don't add new commits to pull requests.

PR #61, opened 2015-06-29
File WRITE library functions and tests added.

gitlarryf:


lib/file.cpp
```diff
@@ -26,6 +26,18 @@ std::string file$bytes(const std::string &filename)
     return r;
 }
 
+void file$writebytes(const std::string &filename, const std::string &data)
+{
+    std::string r;
```
ghewgill: What is this variable for?

lib/file.cpp
```diff
@@ -26,6 +26,18 @@ std::string file$bytes(const std::string &filename)
...
+    std::ofstream f(filename, std::ios::binary);
+    if (not f.is_open()) {
+        throw RtlException("FileOpenError", filename);
+    }
+    if (not f.write((char*)data.c_str(), data.length())) {
```
ghewgill: Do not use a cast, it is not necessary.

lib/file.cpp
```diff
@@ -40,4 +52,21 @@ std::vector<utf8string> file$lines(const std::string &filename)
     return r;
 }
 
+void file$writelines(const std::string &filename, const std::vector<utf8string> &lines)
+{
+    std::string r;
```
ghewgill: Unused variable.

lib/file.cpp
```diff
@@ -40,4 +52,21 @@ std::vector<utf8string> file$lines(const std::string &filename)
...
+    std::ofstream f(filename, std::ios::out | std::ios::trunc);
+    if (not f.is_open()) {
+        throw RtlException("FileOpenError", filename);
+    }
+    for (auto s : lines) {
```
ghewgill: No space before `:`.

lib/file.cpp
```diff
@@ -40,4 +52,21 @@ std::vector<utf8string> file$lines(const std::string &filename)
...
+        if (not s.empty()) {
+            if (not f.write(s.c_str(), s.length())) {
```
ghewgill: Are you sure you want to use `f.write()` here? Why not just `f << s`?

lib/file.cpp
```diff
@@ -40,4 +52,21 @@ std::vector<utf8string> file$lines(const std::string &filename)
...
+                throw RtlException("FileWriteError", filename);
+            }
+        } 
+        f.write((char*)"\n", 1);    // Always write a \n, at the end of each line in the array.
```
ghewgill: Do not use a cast. Also, `f << "\n"`.

lib/file.neon
```diff
@@ -1,13 +1,19 @@
...
 DECLARE EXCEPTION DirectoryExists
 DECLARE EXCEPTION FileOpenError
+DECLARE EXCEPTION FileWriteError
 DECLARE EXCEPTION PathNotFound
 DECLARE EXCEPTION PermissionDenied
 
 DECLARE NATIVE FUNCTION bytes(filename: String): Bytes
 DECLARE NATIVE FUNCTION lines(filename: String): Array<String> 
+
+DECLARE NATIVE FUNCTION writebytes(filename: String, data: Bytes)
```
ghewgill: Functions in alphabetical order please. Blank lines not necessary.

t/file-writebytes.neon
```diff
@@ -0,0 +1,26 @@
+IMPORT file
+
+VAR buf: Bytes
+VAR bytes: Bytes
+
+buf.from_array([4, 3, 2, 0, 1])
+
+TRY
+    file.writebytes("bytetest.bin", buf)
+EXCEPTION file.FileOpenError
```
ghewgill: Do you expect to get this exception? It doesn't appear so. If you aren't specifically testing the exception, then don't attempt to catch it in the test. Not catching an exception in a test will cause the test to fail. (So will writing the "write file open error" that you aren't expecting, but it's best to keep irrelevant stuff out of test cases.)

t/file-writelines.neon
```diff
@@ -0,0 +1,31 @@
+IMPORT file
+
+VAR lines: Array<String> := ["Hello", "Neon", "", "World!"]
+
+TRY
+    file.writelines("linetest.txt", lines)
+EXCEPTION file.FileOpenError
```
ghewgill: Same comments as above here for all the exception catches. You aren't expecting any of them.

t/file-writelines.neon
```diff
@@ -0,0 +1,31 @@
...
+    data := file.lines("linetest.txt")
+EXCEPTION file.FileOpenError
+    print("read file open error")
+END TRY
+
+print("\(data.size()) lines read.")
+%= 4 lines read.
+
+FOREACH s OF data DO
+    print("\(s)")
```
ghewgill: Since `s` is itself a string, interpolation is redundant here and `print(s)` is simpler.

lib/file.cpp
gitlarryf: I did it this way to be able to test the return value, and raise a Neon exception we were not able to write to the file.

lib/file.cpp
```diff
@@ -40,4 +51,20 @@ std::vector<utf8string> file$lines(const std::string &filename)
...
+    if (not f.is_open()) {
+        throw RtlException("FileOpenError", filename);
+    }
+    for (auto s: lines) {
+        if (not s.empty()) {
+            if (not f.write(s.c_str(), s.length())) {
+                throw RtlException("FileWriteError", filename);
+            }
+        }
+		f << "\n";	// Always write a \n, at the end of each line in the array.
```
ghewgill: No tabs please.

lib/file.cpp
```diff
@@ -40,4 +51,21 @@ std::vector<utf8string> file$lines(const std::string &filename)
...
 
+void file$writelines(const std::string &filename, const std::vector<utf8string> &lines)
+{
+    std::ofstream f(filename, std::ios::out | std::ios::trunc);
+    if (not f.is_open()) {
+        throw RtlException("FileOpenError", filename);
+    }
+    for (auto s: lines) {
+        if (not s.empty()) {
+			f << s.c_str();
```
ghewgill: This should be `s.str()`, not `s.c_str()`.

lib/file.cpp
```diff
@@ -40,4 +51,21 @@ std::vector<utf8string> file$lines(const std::string &filename)
...
+			if (f.fail()) {
```
ghewgill: Also, you've got more tabs here again.

lib/file.cpp
```diff
@@ -40,4 +51,21 @@ std::vector<utf8string> file$lines(const std::string &filename)
...
 }
 
+void file$writelines(const std::string &filename, const std::vector<utf8string> &lines)
+{
+    std::ofstream f(filename, std::ios::out | std::ios::trunc);
+    if (not f.is_open()) {
+        throw RtlException("FileOpenError", filename);
+    }
+    for (auto s: lines) {
+        if (not s.empty()) {
```
ghewgill: You know, this would be simplified if you remove the check for `s.empty()`, and do a `f << s.str() << "\n"` for each line. Writing no characters to a file (for blank lines) is not going to be a performance problem.

lib/file.neon
```diff
@@ -1,13 +1,19 @@
...
 DECLARE EXCEPTION DirectoryExists
 DECLARE EXCEPTION FileOpenError
+DECLARE EXCEPTION FileWriteError
 DECLARE EXCEPTION PathNotFound
 DECLARE EXCEPTION PermissionDenied
 
 DECLARE NATIVE FUNCTION bytes(filename: String): Bytes
-DECLARE NATIVE FUNCTION lines(filename: String): Array<String> 
 DECLARE NATIVE FUNCTION mkdir(path: String)
+DECLARE NATIVE FUNCTION lines(filename: String): Array<String>
```
ghewgill: These should be in alphabetical order. Somehow `lines` moved.

lib/file.neon
```diff
@@ -1,13 +1,19 @@
...
 DECLARE EXCEPTION PathNotFound
 DECLARE EXCEPTION PermissionDenied
 
 DECLARE NATIVE FUNCTION bytes(filename: String): Bytes
-DECLARE NATIVE FUNCTION lines(filename: String): Array<String> 
 DECLARE NATIVE FUNCTION mkdir(path: String)
+DECLARE NATIVE FUNCTION lines(filename: String): Array<String> 
+DECLARE NATIVE FUNCTION writebytes(filename: String, data: Bytes)
+DECLARE NATIVE FUNCTION writelines(filename: String, data: Array<String>)
+
```
ghewgill: Please don't leave blank lines at the end of the file.

lib/file.cpp
```diff
@@ -40,4 +51,34 @@ std::vector<utf8string> file$lines(const std::string &filename)
     return r;
 }
 
+void file$appendlines(const std::string &filename, const std::vector<utf8string> &lines)
```
ghewgill: I don't think this function belongs here (as discussed).

lib/file.cpp
```diff
@@ -26,6 +26,17 @@ std::string file$bytes(const std::string &filename)
     return r;
 }
 
+void file$writebytes(const std::string &filename, const std::string &data)
```
ghewgill: This function should be moved down so it's declared in the same order as in the neon file (alphabetical order).

PR #63, opened 2015-07-13
Add file.exists() function

gitlarryf:


lib/file.cpp
```diff
@@ -26,6 +27,12 @@ std::string file$bytes(const std::string &filename)
     return r;
 }
 
+bool file$exists(const std::string &filename)
+{
+    struct stat buffer;
+    return stat(filename.c_str(), &buffer) == 0;
```
ghewgill: For an existence check, use the access() function instead of stat(). We don't need all the extra crap returned by stat().

ghewgill:
Commit messages should be written in the imperative present tense. So just "Add exists() function" instead of "Added". See http://chris.beams.io/posts/git-commit/ for a good read about this.

PR #82, opened 2015-09-15
Fix compare of int variable to size_t variable

gitlarryf:


ghewgill:
Change the commit message so it actually says what is going on here. This  is not a comparison.

PR #85, opened 2015-09-21
Added curses constant values

gitlarryf:


lib/curses.cpp
```diff
@@ -31,6 +31,11 @@ static std::vector<chtype> chtypes_from_numbers(const std::vector<Number> &chstr
 
 namespace rtl {
 
+Number curses$KEY_F(Number c)
```
ghewgill: This doesn't need to be a native function. In the future when the compiler can identify pure functions, it won't be able to do that for native functions. So this should be implemented in neon.

lib/curses.neon
```diff
@@ -194,6 +450,7 @@ DECLARE NATIVE FUNCTION standout()
 DECLARE NATIVE FUNCTION wstandout(win: Window)
 DECLARE NATIVE FUNCTION start_color()
 DECLARE NATIVE FUNCTION stdscr(): Window
+DECLARE NATIVE FUNCTION curscr(): Window
```
ghewgill: To be consistent with the order of functions already in this file, this should be declared right before `curses_version`.

lib/curses.neon
```diff
@@ -12,6 +12,262 @@ DECLARE NATIVE CONST OK: Number
...
+DECLARE NATIVE CONST KEY_ALT_L          :Number % Left-alt
+DECLARE NATIVE CONST KEY_ALT_R          :Number % Right-alt
+DECLARE NATIVE CONST KEY_RESIZE         :Number % Window resize
+DECLARE NATIVE CONST KEY_SUP            :Number % Shifted up arrow
+DECLARE NATIVE CONST KEY_SDOWN          :Number % Shifted down arrow
+
+DECLARE NATIVE CONST KEY_MIN            :Number % Minimum curses key value
+DECLARE NATIVE CONST KEY_MAX            :Number % Maximum curses key
+
+
```
ghewgill: Don't use double blank lines.

lib/curses.neon
```diff
@@ -12,6 +12,262 @@ DECLARE NATIVE CONST OK: Number
...
+DECLARE NATIVE CONST KEY_SREPLACE       :Number % shifted replace key
+DECLARE NATIVE CONST KEY_SRIGHT         :Number % shifted right arrow
+DECLARE NATIVE CONST KEY_SRSUME         :Number % shifted resume key
+DECLARE NATIVE CONST KEY_SSAVE          :Number % shifted save key
+DECLARE NATIVE CONST KEY_SSUSPEND       :Number % shifted suspend key
+DECLARE NATIVE CONST KEY_SUNDO          :Number % shifted undo key
+DECLARE NATIVE CONST KEY_SUSPEND        :Number % suspend key
+DECLARE NATIVE CONST KEY_UNDO           :Number % undo key
+
+% PDCurses-specific key definitions -- PC only
```
ghewgill: What do you intend to do with the key definitions that appear in PDcurses but not in ncurses? Currently those are causing a build failure.

lib/curses.neon
```diff
@@ -12,6 +12,120 @@ DECLARE NATIVE CONST OK: Number
...
+ |  Video Attributes
+ |
+ |%
+
+DECLARE NATIVE CONST A_CHARTEXT         :Number
+
+%|----------------------------------------------------------------------
+ |
+ |  Keyboard, and Key Definitions.
+ |  Many are just for compatibility.
```
ghewgill: For compatibility with what?

lib/curses.neon
```diff
@@ -12,6 +12,120 @@ DECLARE NATIVE CONST OK: Number
...
+ |
+ |%
+
+DECLARE NATIVE CONST KEY_CODE_YES       :Number % If get_wch() gives a key code
```
ghewgill: Please put a space after the colon here (and on all other similar lines).

ghewgill:
Also, all these commits should be collapsed into one.

PR #90, opened 2015-09-26
Curses constants

gitlarryf:


lib/curses.neon
```diff
@@ -12,8 +12,152 @@ DECLARE NATIVE CONST OK: Number
...
+DECLARE NATIVE CONST A_BLINK            : Number
+DECLARE NATIVE CONST A_BOLD             : Number
+DECLARE NATIVE CONST A_ATTRIBUTES       : Number
+DECLARE NATIVE CONST A_COLOR            : Number
+DECLARE NATIVE CONST A_ITALIC           : Number
+DECLARE NATIVE CONST A_PROTECT          : Number
+
+%|----------------------------------------------------------------------
+ |
+ |  Color Macros
```
ghewgill: These aren't macros. Suggest "Color constants" or something.

gitlarryf: Can I leave it as Macro's, but change it to note that they are Macro's handled by functions, or something?  Because most of them ARE in fact function macro's, in the original source, but I'm not married to the concept of keeping the word Macros in this context...

ghewgill: You're commenting the Neon code. Neon doesn't have "macros". You are defining constants. The underlying implementation is not relevant at this level.

PR #106, opened 2015-11-17
Added SID Library

gitlarryf:


ghewgill:
SID module pull requests need to be moved to https://github.com/ghewgill/neon-easysid.

PR #109, opened 2015-12-22
Added RTL exception for substring() function

gitlarryf:


lib/global.neon
```diff
@@ -142,6 +142,13 @@ DECLARE EXCEPTION Utf8Encoding
  |%
 DECLARE EXCEPTION ValueRange
 
+%|
+ |  Exception: PastEndOfString
```
ghewgill: For consistency with other similar exceptions, name this one `StringIndex`.

lib/global.neon
```diff
@@ -142,6 +142,13 @@ DECLARE EXCEPTION Utf8Encoding
...
+ |
+ |  An attempt was made to read a substring, with a length that goes beyond the end of the string.
```
ghewgill: This description is too specific. The exception could be used for things than other than substrings. For consistency with the other (currently existing) exceptions, something like simply:

> A string index is not valid or out of range.

src/intrinsic.cpp
```diff
@@ -85,6 +85,9 @@ std::string global$substring(const std::string &s, Number offset, Number length)
 {
     assert(number_is_integer(offset));
     assert(number_is_integer(length));
+    if (number_to_uint32(offset) + number_to_uint32(length) > s.length()) {
```
ghewgill: This doesn't properly handle overflow of 32-bit unsigned values.

src/intrinsic.cpp
```diff
@@ -85,6 +85,9 @@ std::string global$substring(const std::string &s, Number offset, Number length)
...
+        throw RtlException(Exception_global$PastEndOfString, "length is past the end of string");
```
ghewgill: This exception message doesn't give the caller any additional useful information. At the very least, this message should contain the value of the offending index (see `exec_INDEXAR` for similar examples).

lib/global.neon
```diff
@@ -142,6 +142,14 @@ DECLARE EXCEPTION Utf8Encoding
  |%
 DECLARE EXCEPTION ValueRange
 
+%|
+ |  Exception: StringIndex
```
ghewgill: Make sure you keep the items in this file in the same (overall) alphabetical order.

lib/global.neon
```diff
@@ -142,6 +142,14 @@ DECLARE EXCEPTION Utf8Encoding
...
+ |
+ |  An attempt was made to read or reference a string beyond with a length
```
ghewgill: This is still too specific. See my previous suggestion. Consistency is important.

t/substring.neon
```diff
@@ -0,0 +1,6 @@
+VAR s: String := "Hello, World!"
+print(substring(s, 0, 5))
+%= Hello
+
+print(substring(s, 5, 10))
+%! StringIndex
```
ghewgill: You can't combine %= and %! like this in a single file. If you want to catch a runtime exception as well as match output from previous expression, you'll have to use `TRY`.

src/intrinsic.cpp
```diff
@@ -85,6 +85,9 @@ std::string global$substring(const std::string &s, Number offset, Number length)
 {
     assert(number_is_integer(offset));
     assert(number_is_integer(length));
+    if (number_to_uint32(offset) + number_to_uint32(length) > s.length()) {
+        throw RtlException(Exception_global$StringIndex, "length is past the end of string");
```
ghewgill: What happened to my previous suggestion for this message?

PR #112, opened 2016-01-02
Added OTHERS keyword for WHEN clauses

gitlarryf:


src/parser.cpp
```diff
@@ -1209,16 +1215,16 @@ const Statement *Parser::parseCaseStatement()
...
         ++i;
         while (tokens[i].type != END) {
             const Statement *stmt = parseStatement();
             if (stmt != nullptr) {
-                else_statements.push_back(stmt);
+                others_statements.push_back(stmt);
             }
         }
-    }
+    }
```
ghewgill: Please make sure you don't make unrelated changes, especially trailing spaces.

t/case.neon
```diff
@@ -1,6 +1,5 @@
-VAR a: Number
-VAR s: String
-a := 0
+VAR a: Number := 0
+VAR s: String := ""
```
ghewgill: Why did you change this initialization? `a` is fine, but `s` was *specifically* uninitialized. Please keep it that way.

tools/helium.py
```diff
@@ -1469,6 +1470,9 @@ def parse_case_statement(self):
             conditions = []
             while True:
                 t = self.tokens[self.i]
+                if t is OTHERS:
```
ghewgill: Same comments about the parsing here applies as in `parser.cpp`. I think doing it the way I suggest will be easier.

gh-pages/docs/index.md
```diff
@@ -171,6 +171,7 @@ The following keywords are predefined and may not be used for any other purpose.
 | <a href="#keyword-NOT">`NOT`</a> | logical negation |
 | <a href="#keyword-OR">`OR`</a> | logical disjunction |
 | <a href="#keyword-OUT">`OUT`</a> | function parameter passing mode |
+| <a href="#keyword-OTHERS">`OTHERS`</a> | alternative condition in `WHEN` statements |
```
ghewgill: This is a CASE statement, not a WHEN statement.

gh-pages/docs/index.md
```diff
@@ -686,7 +687,7 @@ The possible kinds of `WHEN` clauses are:
 
 More than one of the above forms may be included in a `WHEN` clause, separated by commas.
 The values of `WHEN` clauses must not overlap.
-The optional `ELSE` clause is executed when none of the `WHEN` clauses match.
+The optional `OTHERS` clause is executed when no other `WHEN` clauses match.
```
ghewgill: Say "The optional `WHEN OTHERS` clause"

src/parser.cpp
```diff
@@ -1201,7 +1201,7 @@ const Statement *Parser::parseCaseStatement()
         }
         ++i;
         std::vector<const Statement *> statements;
-        while (tokens[i].type != WHEN && tokens[i].type != ELSE && tokens[i].type != END && tokens[i].type != END_OF_FILE) {
+        while (tokens[i].type != WHEN && tokens[i].type != END && tokens[i].type != END_OF_FILE && tokens[i+1].type != OTHERS) {
```
ghewgill: Why are you checking for `OTHERS` here? That isn't necessary. A when clause ends with another `WHEN`, or an `END` (no longer with an `ELSE`, so all you should need to do is remove that).

tools/helium.py
```diff
@@ -1487,21 +1488,21 @@ def parse_case_statement(self):
                 self.i += 1
             self.expect(DO)
             statements = []
-            while self.tokens[self.i] is not WHEN and self.tokens[self.i] is not ELSE and self.tokens[self.i] is not END and self.tokens[self.i] is not END_OF_FILE:
+            while self.tokens[self.i] is not WHEN and self.tokens[self.i] is not OTHERS and self.tokens[self.i] is not END and self.tokens[self.i] is not END_OF_FILE:
```
ghewgill: You've changed this differently from how you changed the C++ version. They're both wrong. Make this one match the fixed C++ one.

PR #114, opened 2016-01-04
added binary encoding/decoding lib

gitlarryf:


lib/encoding.neon
```diff
@@ -0,0 +1,159 @@
...
+ |  File: encoding
+ |
+ |  Functions for encoding and decoding ASCII or Binary data in various formats.
+ |%
+IMPORT bitwise
+
+EXPORT base64Encode
+EXPORT base64Decode
+
+TYPE Mode IS ENUM
```
ghewgill: I'm not sure having an enum for this is terribly helpful. First, terminology - it's not a "mode", but a "state" (as in "state machine"). Second, this would be equivalently served by a simple number. Third, you don't actually even have to keep track since in your loop you already know what offset you're working with, so you can tell what you're doing with just a MOD 3.

lib/encoding.neon
```diff
@@ -0,0 +1,159 @@
...
+    BYTE1
+    BYTE2
+    BYTE3
+    BYTE4
+END ENUM
+
+CONSTANT Base64EncodeTable: String := "ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789+/"
+
+CONSTANT Base64DecodeTable: Array<Number> := [
+    255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255,    % 0x00 - 0x0f
```
ghewgill: There's no reason to store the value "255" in here, since it's numbers, not bytes. Use -1 for a flag saying the value is not valid (and that's much easier to check later).

lib/encoding.neon
```diff
@@ -0,0 +1,159 @@
...
+ |  Encodes the provided string into Base64 encoded data.
+ |
+ |  Parameters:
+ |      src - The source data to encode.
+ |
+ |%
+FUNCTION base64Encode(src: Bytes): String
+    VAR i: Number := 0
+    VAR encoded: String := ""
+    VAR char1, char2: Number
```
ghewgill: You don't actually need two variables here, because you never use both `char1` and `char2` at the same time. Just one would suffice.

lib/encoding.neon
```diff
@@ -0,0 +1,159 @@
...
+    VAR status: Mode := Mode.BYTE1
+    
+    LET source: Array<Number> := src.toArray()
+
+    WHILE i < src.size() DO
+        VAR char: Number := source[i]
```
ghewgill: Always use `LET` where possible.

lib/encoding.neon
```diff
@@ -0,0 +1,159 @@
...
+        CASE status
+            WHEN Mode.BYTE1 DO
+                encoded.append(Base64EncodeTable[bitwise.shiftRight32(char, 2)])
+                char1 := bitwise.and32(bitwise.shiftLeft32(char, 4), 0x30)
```
ghewgill: It might be more illustrative to use binary `0b00110000` instead of hex `0x30` here. You can even do `0b0011_0000` for extra clarity.

lib/encoding.neon
```diff
@@ -0,0 +1,159 @@
...
+ |%
+FUNCTION base64Encode(src: Bytes): String
+    VAR i: Number := 0
+    VAR encoded: String := ""
+    VAR char1, char2: Number
+    VAR status: Mode := Mode.BYTE1
+    
+    LET source: Array<Number> := src.toArray()
+
+    WHILE i < src.size() DO
```
ghewgill: Consider using a `FOR` loop instead of `WHILE` here. You already know ahead of time exactly how many iterations you want to run, and a `FOR` loop is appropriate for that. Also, using a `FOR` loop avoids reevaluating `src.size()` every time.

lib/encoding.neon
```diff
@@ -0,0 +1,159 @@
...
+                status := Mode.BYTE3
+            WHEN Mode.BYTE3 DO
+                encoded.append(Base64EncodeTable[bitwise.or32(char2, bitwise.shiftRight32(char, 6))])
+                encoded.append(Base64EncodeTable[bitwise.and32(char, 0x3F)])
+                status := Mode.BYTE1
+        END CASE
+        INC i
+    END WHILE
+    
+    CASE (status)
```
ghewgill: No need for parentheses here.

lib/encoding.neon
```diff
@@ -0,0 +1,159 @@
...
+ |%
+FUNCTION base64Decode(src: String): Bytes
+    VAR i: Number := 0
+    VAR char1, char2, char3: Number := 0
+    VAR decoded: Array<Number> := []
+    VAR mode: Mode := Mode.BYTE1
+
+    WHILE i < src.length() DO
+        VAR char: Number := Base64DecodeTable[ord(src[i])]
+        IF bitwise.and32(char, 0x80) > 0 THEN
```
ghewgill: Using a sentinel value of -1 as above simplifies this test (so you don't have to call bitwise.and32).

lib/encoding.neon
```diff
@@ -0,0 +1,159 @@
...
+            INC i
+            NEXT WHILE % Eat chars that are not part of the Base64 alphabet, include \r and \n.
+        END IF
+        IF char = 0x40 THEN
+            EXIT WHILE % We're at the end, so bail out.
```
ghewgill: ...the end of what? What is special about the `@` character here?

lib/encoding.neon
```diff
@@ -0,0 +1,159 @@
...
+                decoded.append(bitwise.and32(bitwise.or32(bitwise.shiftLeft32(char3, 6), char), 0xFF))
+                mode := Mode.BYTE1
+        END CASE
+        INC i
+    END WHILE
+
+    RETURN decoded.toBytes()
+END FUNCTION
+
+FUNCTION stringToBytes(s: String): Bytes
```
ghewgill: Everything from here to the end of the file should be in a `t/encoding-test.neon` file, just like everything else in the standard library.

lib/encoding.neon
```diff
@@ -0,0 +1,159 @@
...
+    RETURN s.toBytes()
+END FUNCTION
+
+BEGIN MAIN
+    print("Running Base64Encode tests...")
```
ghewgill: Do not print out messages. That's done by the test framework after you move this to a separate test file.

lib/encoding.neon
```diff
@@ -0,0 +1,159 @@
...
+ |
+ |  Functions for encoding and decoding ASCII or Binary data in various formats.
+ |%
+IMPORT bitwise
+
+EXPORT base64Encode
+EXPORT base64Decode
+
+TYPE Mode IS ENUM
+    BYTE1
```
ghewgill: Also, as a matter of style, don't use all uppercase here (even though you can in this case because it contains a digit). Neon style is to use lower case for enum values.

lib/encoding.neon
```diff
@@ -0,0 +1,132 @@
...
+ |
+ |  Functions for encoding and decoding ASCII or Binary data in various formats.
+ |%
+IMPORT bitwise
+
+EXPORT base64Encode
+EXPORT base64Decode
+
+TYPE State IS ENUM
+    BYTE1
```
ghewgill: Don't use all uppercase for these. Neon style is to use lower case. `byte1`, `byte2`, etc.

lib/encoding.neon
```diff
@@ -0,0 +1,132 @@
...
+                status := State.BYTE3
+            WHEN State.BYTE3 DO
+                encoded.append(Base64EncodeTable[bitwise.or32(char2, bitwise.shiftRight32(char, 6))])
+                encoded.append(Base64EncodeTable[bitwise.and32(char, 0b00111111)])
+                status := State.BYTE1
+        END CASE
+    END FOR
+
+    CASE status
+        % In base64, when the data being encoded is NOT a multiple of 6 bits, needs to be padded.  The following pads the output data
```
ghewgill: Line too long. Split it at something reasonable (I usually try to split comments at about 80).

lib/encoding.neon
```diff
@@ -0,0 +1,132 @@
...
+        % and signals we've hit the end of the encoded data when decoding...
```
ghewgill: Why does this comment end with `...`? Did you have another thought to add here? When a reader encounters this, will they wonder what else you were going to say but didn't? Comments should be complete sentences, with a period at the end.

lib/encoding.neon
```diff
@@ -0,0 +1,132 @@
...
+ |  Parameters:
+ |      src - The source Base64 to decode.
+ |
+ |%
+FUNCTION base64Decode(src: String): Bytes
+    VAR char1, char2, char3: Number := 0
+    VAR decoded: Array<Number> := []
+    VAR state: State := State.BYTE1
+
+    FOR i:= 0 TO src.length() -1 DO
```
ghewgill: Spaces around `:=`.

lib/encoding.neon
```diff
@@ -0,0 +1,132 @@
...
+ |
+ |%
+FUNCTION base64Encode(src: Bytes): String
+    VAR encoded: String := ""
+    VAR char1, char2: Number
+    VAR status: State := State.BYTE1
+
+    LET source: Array<Number> := src.toArray()
+
+    FOR i:= 0 TO src.size()-1 DO
```
ghewgill: Spaces around `:=`.

lib/encoding.neon
```diff
@@ -0,0 +1,132 @@
...
+ |
+ |%
+FUNCTION base64Encode(src: Bytes): String
+    VAR encoded: String := ""
+    VAR char1, char2: Number
+    VAR status: State := State.byte1
+
+    LET source: Array<Number> := src.toArray()
+
+    FOR i:= 0 TO src.size()-1 DO
```
ghewgill: Spaces around `:=`.

lib/encoding.neon
```diff
@@ -0,0 +1,132 @@
...
+ |  Parameters:
+ |      src - The source Base64 to decode.
+ |
+ |%
+FUNCTION base64Decode(src: String): Bytes
+    VAR char1, char2, char3: Number := 0
+    VAR decoded: Array<Number> := []
+    VAR state: State := State.byte1
+
+    FOR i:= 0 TO src.length() -1 DO
```
ghewgill: Spaces around `:=`. Also, be consistent about spaces around `-`. This is different from the loop in the encoder.

lib/encoding.neon
```diff
@@ -0,0 +1,132 @@
...
+        LET char: Number := Base64DecodeTable[ord(src[i])]
+        IF char = -1 THEN
+            NEXT FOR % Eat chars that are not part of the Base64 alphabet, including \r and \n, and whitespace.
+        END IF
+        IF char = 0x40 THEN
+            % If src[i] is the ASCII "=", then we've hit the end of the ACTUAL base64 encoded data, so we can break out now.  The value stored in
```
ghewgill: Line too long again.

lib/encoding.neon
```diff
@@ -0,0 +1,132 @@
...
+    byte1
+    byte2
+    byte3
+    byte4
+END ENUM
+
+CONSTANT Base64EncodeTable: String := "ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789+/"
+
+LET Base64DecodeTable: Array<Number> := [
+     -1,  -1,  -1,  -1,  -1,  -1,  -1,  -1,  -1,  -1, -1,  -1,  -1,  -1,  -1,  -1,    % 0x00 - 0x0f
```
ghewgill: There's inconsistent spacing in this table. Column 11 has one fewer leading space than the rest of the columns.

lib/encoding.neon
```diff
@@ -0,0 +1,132 @@
...
+     -1,  -1,  -1,  -1,  -1,  -1,  -1,  -1,  -1,  -1, -1,  -1,  -1,  -1,  -1,  -1,    % 0x90 - 0x9f
+     -1,  -1,  -1,  -1,  -1,  -1,  -1,  -1,  -1,  -1, -1,  -1,  -1,  -1,  -1,  -1,    % 0xa0 - 0xaf
+     -1,  -1,  -1,  -1,  -1,  -1,  -1,  -1,  -1,  -1, -1,  -1,  -1,  -1,  -1,  -1,    % 0xb0 - 0xbf
+     -1,  -1,  -1,  -1,  -1,  -1,  -1,  -1,  -1,  -1, -1,  -1,  -1,  -1,  -1,  -1,    % 0xc0 - 0xcf
+     -1,  -1,  -1,  -1,  -1,  -1,  -1,  -1,  -1,  -1, -1,  -1,  -1,  -1,  -1,  -1,    % 0xd0 - 0xdf
+     -1,  -1,  -1,  -1,  -1,  -1,  -1,  -1,  -1,  -1, -1,  -1,  -1,  -1,  -1,  -1,    % 0xe0 - 0xef
+     -1,  -1,  -1,  -1,  -1,  -1,  -1,  -1,  -1,  -1, -1,  -1,  -1,  -1,  -1,  -1,    % 0xf0 - 0xff
+]
+
+%FUNCTION base64Encode(src: Bytes): String
```
ghewgill: What's this line for? Why is there a commented-out function header here?

t/encoding-base64.neon
```diff
@@ -0,0 +1,61 @@
+IMPORT encoding
+
+FUNCTION stringToBytes(s: String): Bytes
+    RETURN s.toBytes()
+END FUNCTION
+
+FUNCTION t(b: Bytes)
+    ASSERT (b = encoding.base64Decode(encoding.base64Encode(b)))
```
ghewgill: ASSERT needs no parentheses around its test.

t/encoding-base64.neon
```diff
@@ -0,0 +1,61 @@
...
+END FUNCTION
+
+% First, perform encoding tests using the RFC examples...
```
ghewgill: You're using `...` again.

lib/encoding.neon
```diff
@@ -0,0 +1,132 @@
...
+ |%
+FUNCTION base64Decode(src: String): Bytes
+    VAR char1, char2, char3: Number := 0
+    VAR decoded: Array<Number> := []
+    VAR state: State := State.byte1
+
+    FOR i:= 0 TO src.length() -1 DO
+        LET char: Number := Base64DecodeTable[ord(src[i])]
+        IF char = -1 THEN
+            NEXT FOR % Eat chars that are not part of the Base64 alphabet, including \r and \n, and whitespace.
```
ghewgill: Although you can put a comment anywhere, it's more readable to put the comment on its own line:

    % Eat chars that are not part of the Base64 alphabet, including \r and \n, and whitespace.
    NEXT FOR

lib/encoding.neon
```diff
@@ -0,0 +1,132 @@
...
+        END IF
+        IF char = 0x40 THEN
+            % If src[i] is the ASCII "=", then we've hit the end of the ACTUAL base64 encoded data, so we can break out now.  The value stored in
+            % this cell, is actuall 0x40, which is the ASCII "@" (AT-Symbol), so at first glance, this might look a little strange, but this is
```
ghewgill: Wait, what? Why are you checking the value of `char` against a sentinel value, instead of directly checking whether `src[i] = "="`? Using a direct check against equals would save a lot of explanation in the comments here.

lib/encoding.neon
```diff
@@ -0,0 +1,135 @@
...
+    byte4
+END ENUM
+
+CONSTANT Base64EncodeTable: String := "ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789+/"
+
+LET Base64DecodeTable: Array<Number> := [
+     -1,  -1,  -1,  -1,  -1,  -1,  -1,  -1,  -1,  -1,  -1,  -1,  -1,  -1,  -1,  -1,    % 0x00 - 0x0f
+     -1,  -1,  -1,  -1,  -1,  -1,  -1,  -1,  -1,  -1,  -1,  -1,  -1,  -1,  -1,  -1,    % 0x10 - 0x1f
+     -1,  -1,  -1,  -1,  -1,  -1,  -1,  -1,  -1,  -1,  -1,  62,  -1,  -1,  -1,  63,    % 0x20 - 0x2f
+     52,  53,  54,  55,  56,  57,  58,  59,  60,  61,  -1,  -1,  -1,  64,  -1,  -1,    % 0x30 - 0x3f
```
ghewgill: Still have the value 64 in here which no longer has a purpose.

lib/encoding.neon
```diff
@@ -0,0 +1,135 @@
...
+ |
+ |%
+FUNCTION base64Encode(src: Bytes): String
+    VAR encoded: String := ""
+    VAR char1, char2: Number
+    VAR status: State := State.byte1
+
+    LET source: Array<Number> := src.toArray()
+
+    FOR i := 0 TO src.size() - 1 DO
```
ghewgill: Actually, you could do this with a `FOREACH` loop now.

lib/encoding.neon
```diff
@@ -0,0 +1,135 @@
...
+ |  Parameters:
+ |      src - The source Base64 to decode.
+ |
+ |%
+FUNCTION base64Decode(src: String): Bytes
+    VAR char1, char2, char3: Number := 0
+    VAR decoded: Array<Number> := []
+    VAR state: State := State.byte1
+
+    FOR i := 0 TO src.length() - 1 DO
```
ghewgill: And a `FOREACH` here too.

t/encoding-base64.neon
```diff
@@ -0,0 +1,61 @@
...
+ASSERT encoding.base64Decode("Zm9vYmE=").toString() = "fooba"
+ASSERT encoding.base64Decode("Zm9vYmFy").toString() = "foobar"
+
+% Perform minimal "fuzz" tests...
+ASSERT encoding.base64Decode("Zm9vYmFy\r   \nZm9vYmFy").toString() = "foobarfoobar"
+ASSERT encoding.base64Decode("Zm9vYmFyDQ  pmb29iYXI=").toString() = "foobar\r\nfoobar"
+ASSERT encoding.base64Decode("!@#$%^&*)(_+_)==").toString() = ""
+ASSERT encoding.base64Decode(encoding.base64Encode(stringToBytes("foobar"))).toString() = "foobar"
+
+% Most thourogh tests of all, test every bit, at every position across all 3 base64 Bytes.
```
ghewgill: Spelling: thorough.

PR #126, opened 2016-02-09
Add AppVeyor CI Support

gitlarryf:


README.md
```diff
@@ -30,7 +30,7 @@ To build Neon, the following are required:
 - [SCons](http://scons.org)
 - A C++11 compiler (a modern `gcc`, `clang`, or Visual Studio 2013 or later)
 
-Current master branch build status: [![Build Status](https://travis-ci.org/ghewgill/neon-lang.svg?branch=master)](https://travis-ci.org/ghewgill/neon-lang)
+Current master branch build status: [![Build Status](https://travis-ci.org/ghewgill/neon-lang.svg?branch=master)](https://travis-ci.org/ghewgill/neon-lang) (https://ci.appveyor.com/api/projects/status/github/ghewgill/neon-lang?branch=master&svg=true)
```
ghewgill: This isn't right. It needs to be a button, just like the travis one. You can see the result at [your appveyor branch](https://github.com/gitlarryf/neon-lang/tree/appveyor), which clearly doesn't show a button here.

appveyor.yml
```diff
@@ -0,0 +1,16 @@
+clone_folder: C:\projects\neon-lang\
```
ghewgill: This shouldn't be necessary. If you can, remove this line and the `--directory=` option to scons (unless there is a good reason that this *needs* to be explicitly specified).

gitlarryf: There is a VERY good reason to have it there, as that is where our NEON-LANG project on AppVeyor is setup to drop our files into...  So, it's either we do a CD, or I pass --directory, which would you prefer?  I'd prefer the --directory, since it automagically tells SCons which folder to go to...

ghewgill: According to https://www.appveyor.com/docs/build-configuration#clone-directory the clone directory default is indeed `c:\projects\whatever`. The next paragraph says *if* it's required by your project, then you can explicitly specify the clone directory. Neon should not have any absolute path references, therefore it is not necessary to ask for a specific clone directory.

gitlarryf: Ok, that is fair enough... I forgot that Appveyor defaults to c:\projects\<project>... But, that doesn't change the SCons calls, right?  I mean, it might pull the files INTO c:\projects\neon-lang, but that does NOT mean that it makes it the current directory...  Hence, the need for the --directory parameter, *OR* the CD c:\projects\neon-lang...  Which is why I wanted the paths "hard-coded" in the *BUILD SYSTEM* so that we were ALWAYS sure where it went to...

PR #142, opened 2016-07-14
Add isDebug() function to runtime library

gitlarryf:


ghewgill:
There's no description here about why a neon program might need to know this information. At this point, this just seems like a leaky abstraction and I don't see any reason to merge it.

gitlarryf:
I thought it might be useful to the Neon programmer who might want to enable extra debugging measures during development, but then have them disabled when running against a release mode runtime.  This prevents the user from having to create his own debug boolean variable that he has to purposely disable before shipping it out to the world, perhaps.

ghewgill:
A "debug" mode executor and a debug flag in a neon program are orthogonal concerns. They don't really relate to one another at all. 

Also consider those users who might install neon from a precompiled package - there won't *be* a debug mode executor.

PR #164, opened 2016-10-10
Add sample program isprime

gitlarryf:


samples/isprime/isprime.neon
```diff
@@ -0,0 +1,184 @@
...
+   fails, we know that n is composite, and describe the base a as 
+   a witness to the compositeness of n. The crucial extra information 
+   we can derive is that if (odd) n is composite, then at least 3/4 
+   of the numbers a with 1 < a < n - 1 are witness to this fact.
+   First calculate m, in Neon we do this by simply shifting the bits
+   right k times, until m is odd and less than n.
+   Next, we calculate a, such that a is less than n - 1, and a > 1.
+   Then, we calculate a ^ m mod n.  If this result is 1, then
+   we need to generate a new "a" and try again.  After iIterations, we
+   can say that n is prime with 1/1024 margin of error.
```
ghewgill: What parameter establishes the 1/1024 margin of error?

samples/isprime/isprime.neon
```diff
@@ -0,0 +1,184 @@
...
+                                    67,  71,  73,  79,  83,  89,  97, 101, 103, 107, 109, 113, 127, 131, 137, 139, 149, 151, 
+                                   157, 163, 167, 173, 179, 181, 191, 193, 197, 199, 211, 223, 227, 229, 233, 239, 241, 251, 257 ]
+
+FUNCTION SizeOf(obj: Array<Number>): Number
+    RETURN obj.size()
+END FUNCTION
+                            
+CONSTANT TestIterations: Number := 5
+
+FUNCTION GetRandomA(OUT a:Number, n:Number)
```
ghewgill: This should be a function returning a number, instead of having a single OUT parameter and no return value.

samples/isprime/isprime.neon
```diff
@@ -0,0 +1,184 @@
...
+    VAR rnd: Number := random.uint32()
+    a := rnd MOD (n - 2) + 1
+END FUNCTION
+
+FUNCTION isOdd(n: Number): Boolean
+    IF binary.get32(n, 0) = TRUE THEN
```
ghewgill: The use of `= TRUE` here is redundant. The `IF` statement tests a boolean value, and that's what `binary.get32()` returns.

samples/isprime/isprime.neon
```diff
@@ -0,0 +1,184 @@
...
+        RETURN TRUE
```
ghewgill: The use of an `IF` statement here at all is redundant. If you want to return a value, just do `RETURN binary.get32(n, 0)`. That's it. No `= TRUE`, no `IF` statement.

samples/isprime/isprime.neon
```diff
@@ -0,0 +1,184 @@
...
+
+    IF NOT isOdd(n) THEN
+        % Rule: The number n is an EVEN number, and therefore CANNOT be prime.
+        IF n = 2 THEN
+            % The number 2 is the only exception to the rule!
+            % Since 2 is the ONLY even number that is prime, we handle this
+            % case a little differently.
+            RETURN TRUE
+        END IF
+        RETURN isprime
```
ghewgill: It would be clearer here to `RETURN FALSE`, so the reader doesn't have to look at `isprime` and figure out whether its value had changed since it was initialized.

samples/isprime/isprime.neon
```diff
@@ -0,0 +1,184 @@
...
+    END IF
+
+    IF m > 0 THEN
+        WHILE binary.get32(m, 0) = TRUE DO
```
ghewgill: Again, the redundant `= TRUE` antipattern.

samples/isprime/isprime.neon
```diff
@@ -0,0 +1,184 @@
...
+            DEC i
+        END IF
+        FOR x := 0 TO k DO
+            r := (r*r) MOD n
+            IF r = (n - 1) THEN
+                isprime := TRUE
+                EXIT FOR
+            END IF
+        END FOR
+        IF NOT isprime THEN
```
ghewgill: Where would `isprime` ever get set to false in this loop? It looks like this whole loop tests the first iteration, setting isprime to true if it passes that first test, then keeps running the rest of the iterations which cannot possibly have any effect on the ultimate outcome.

samples/isprime/isprime.neon
```diff
@@ -0,0 +1,184 @@
...
+            EXIT REPEAT
+        END IF
+        DEC i
+    UNTIL i # 0
+    RETURN isprime
+END FUNCTION
+
+FUNCTION SmallFactor(n: Number): Boolean
+    FOR i := 0 TO SizeOf(SMALL_PRIME)-1 DO
```
ghewgill: Use a `FOREACH` loop here. Then you can get rid of the `SizeOf` function too.

samples/isprime/isprime.neon
```diff
@@ -0,0 +1,184 @@
...
+        IF (n MOD SMALL_PRIME[i] = 0) THEN
+            RETURN FALSE
+        END IF
+    END FOR
+    RETURN TRUE
+END FUNCTION
+
+BEGIN MAIN
+    IF sys.args.size() <= 1 THEN
+        print("\nUsage: \(sys.args[0]) {Number | s-e | s- | -e}")
```
ghewgill: Do not use `\n` inside print statements like this. It's much clearer, if you want a blank line, to have a separate statement that prints a blank line.

samples/isprime/isprime.neon
```diff
@@ -0,0 +1,184 @@
...
+        print(" Where: s is starting integer, and e is ending integer.\n")
+        sys.exit(1)
+    END IF
+
+    VAR sval: Number := 0
+    VAR eval: Number := 0
+    VAR hasDash: Number := string.find(sys.args[1], "-")
```
ghewgill: Consider using `string.split()` here to simplify the logic.

samples/isprime/isprime.neon
```diff
@@ -0,0 +1,184 @@
...
+                IF isOdd(sval) THEN
+                    sval := sval + 2
+                ELSE
+                    INC sval
+                END IF
+            END WHILE
+            print("\(sval) is the next prime.")
+        END IF
+    ELSIF hasDash > 0 THEN
+        % Print pimes within a RANGE...
```
ghewgill: Typo.

samples/isprime/isprime.neon
```diff
@@ -0,0 +1,184 @@
...
+        IF hasDash > 0 THEN
+            sval := num(substring(sys.args[1], 0, hasDash))
+            eval := num(substring(sys.args[1], hasDash+1, sys.args[1].length()-hasDash-1))
+            print("Locating all prime number between \(sval) and \(eval)...")
```
ghewgill: Typo.

samples/isprime/isprime.neon
```diff
@@ -0,0 +1,184 @@
...
+        sval := num(substring(sys.args[1], hasDash+1, sys.args[1].length()-hasDash-1))
+        print("Locating last prime number before \(sval)...")
+        VAR val: Number := 0
+        FOR x := sval TO eval STEP -1 DO
+            val := x
+            IF RabinMiller(val, TestIterations) THEN
+                print("\(val)")
+                EXIT FOR
+            END IF
+            IF isOdd(val) THEN
```
ghewgill: What's the purpose of modifying `val` here when you're just going to reset its value from `x` at the start of the next loop?

samples/isprime/isprime.neon
```diff
@@ -0,0 +1,184 @@
...
+    ELSIF string.hasSuffix(sys.args[1], "-") THEN
+        % Print the next prime AFTER n.
+        IF hasDash > 0 THEN
+            sval := num(substring(sys.args[1], 0, hasDash))
+            print("Locating the next prime number after \(sval)...")
+            WHILE TRUE DO
+                IF RabinMiller(sval, TestIterations) THEN
+                    EXIT WHILE
+                END IF
+                IF isOdd(sval) THEN
```
ghewgill: This won't work if you ask for the primes from "1-", since 1 is odd so you'll increment `sval` to 3 and skip 2 (which is prime).

samples/isprime/isprime.neon
```diff
@@ -0,0 +1,184 @@
...
+    ELSE
+        sval := num(sys.args[1])
+        IF RabinMiller(sval, TestIterations) THEN
+            print("\(sval) is prime.")
+        ELSE
+            print("\(sval) is not prime.")
+        END IF
+    END IF
+END MAIN
+
```
ghewgill: Include a newline at the end of the file.

gitlarryf: There isn't a new line there?  What is line 184?

samples/isprime/isprime.neon
```diff
@@ -0,0 +1,211 @@
...
+    IF m > 0 THEN
+        WHILE binary.get32(m, 0) DO
+            INC k
+            m := binary.shiftRight32(m, 1)
+        END WHILE
+    END IF
+
+    VAR r: Number
+    FOR iCount := 0 TO iIterations LABEL WitnessLoop DO
```
ghewgill: Do you actually want to run this loop `iIterations` times, or `iIterations+1` times? Because right now you're running it `iIterations+1` times.

samples/isprime/isprime.neon
```diff
@@ -0,0 +1,211 @@
...
+        a := GetRandomA(n)
+        r := 1
+        FOR x := 0 TO m - 1 DO
```
ghewgill: If you just want to do something `m` times, it's probably clearer to write `FOR x := 1 TO m`. Saves a subtraction too.

samples/isprime/isprime.neon
```diff
@@ -0,0 +1,211 @@
...
+            r := ((r * a) MOD n)
+        END FOR
+        IF r = 1 OR r = (n - 1) THEN
+            a := GetRandomA(n)
+            NEXT WitnessLoop
+        END IF
+        FOR x := 0 TO k DO
```
ghewgill: Again, do you want to do this loop `k` times, or `k+1` times? It's unclear to me, particularly in light of previous loops.

samples/isprime/isprime.neon
```diff
@@ -0,0 +1,211 @@
...
+            IF r = (n - 1) THEN
+                NEXT WitnessLoop
+            END IF
+        END FOR
+        RETURN FALSE
+    END FOR
+    RETURN TRUE
+END FUNCTION
+
+FUNCTION SmallFactor(n: Number): Boolean
```
ghewgill: What's the return value of this mean? Intuitively, I would guess that `TRUE` means there *is* a small factor, but that's the opposite of what actually happens.

samples/isprime/isprime.neon
```diff
@@ -0,0 +1,211 @@
...
+                ELSE
+                    INC val
+                END IF
+            END FOR
+        END IF
+    ELSE
+        sval := num(nums)
+        output("Testing \(sval) for primality.")
+        IF RabinMiller(sval, TestIterations) THEN
+            print("\(sval) is prime.")
```
gitlarryf: No, only because this is a different test, this directly tests a number for prime, while the others search for a prime.  This can be argued, however, that in quiet mode, it JUST output the number, but not quiet mode it outputs what I have here.  I was torn by this when I wrote it.

samples/isprime/isprime.neon
```diff
@@ -0,0 +1,233 @@
...
+        IF bPar THEN
+            bPar := NOT bPar
+            NEXT FOREACH
+        END IF
+        IF p = "-q" OR p = "-Q" THEN
+            bQuiet := TRUE
+        ELSIF p = "-k" OR p = "-K" THEN
+            suppliedNumTests := TRUE
+            bPar := suppliedNumTests
+            TestIterations := num(sys.args[i+1])
```
ghewgill: This seems to go slightly haywire if you specify a non-numeric argument here. Like `-k a`. Perhaps `num()` should raise an exception for input that's not in fact a number.

gitlarryf: I don't think num() should raise an exception for non-numerical input.  There may be times when you WANT to convert a string to a number, and NOT have it exception on you if there are non-convertible chars in the string.  However, I'm of the opinion that a string of: 123x45 when passed through num(), should return 123, rather than 0.

samples/isprime/isprime.neon
```diff
@@ -0,0 +1,233 @@
...
+            IF TestIterations < 1 THEN
+                Usage()
+            END IF
+        ELSE
+            nums.append(sys.args[i])
```
gitlarryf: I can't do that.  Your earlier review comment says to change the loop to a `WHILE` loop, so `p` is going to go away, but I get your meaning here.

ghewgill: Sure, I just happened to notice this first before the comment about the overall loop structure.

samples/isprime/isprime.neon
ghewgill: I disagree that a function like `num()` should be designed to silently discard data that did not happen to match what it was expecting. Silent failure is the worst kind of failure, because to avoid the problem you have to first validate the data before passing the data to the function. That makes the API error-prone (if you forget to validate) and hard to use (because you have to do two things to call one function).

ghewgill:
Closed pull request due to no activity.

PR #166, opened 2016-12-27
Add source file validation to compiler.

gitlarryf:


src/neonc.cpp
```diff
@@ -47,8 +47,13 @@ int main(int argc, char *argv[])
             if (i != std::string::npos) {
                 source_path = name.substr(0, i+1);
             }
-            std::cout << "Compiling " << name << "...\n";
```
ghewgill: This message should remain *before* any error detection.

src/neonc.cpp
```diff
@@ -47,8 +47,13 @@ int main(int argc, char *argv[])
...
             std::ifstream inf(name);
+            // TODO: Check for .neon file extension?  Maybe. 
+            if (not inf.good()) {
+                std::cout << "\"" << name << "\"" << " not found or is not a neon source file.\n";
```
ghewgill: This check doesn't say anything about whether the file is a neon source file or not. After all, at this point any file can be opened, not just neon source files.

ghewgill: Also, "not found" is not the only error that can possibly happen here. For example, what about "permission denied"? In that case the file *would* exist, but couldn't be opened. Suggest getting a proper error string message.

ghewgill: Also, surrounding the filename with double quotes here and not on the "Compiling ..." line is inconsistent.

PR #179, opened 2017-07-14
Fix module location issue on Windows

gitlarryf:
Fixes issues trying to locate extension modules when more than one path separator is used in one of the Neon Paths.

### DIFF
--- a/src/exec.cpp
+++ b/src/exec.cpp
@@ -63,6 +63,9 @@ std::string just_path(const std::string &name)
     if (i == std::string::npos) {
         i = name.rfind('/');
     }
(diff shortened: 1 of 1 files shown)


src/exec.cpp
```diff
@@ -63,6 +63,9 @@ std::string just_path(const std::string &name)
     if (i == std::string::npos) {
         i = name.rfind('/');
     }
+    if (name.substr(0, i+1).rfind('/')) {
```
ghewgill: This doesn't work because the value returned when the character is not found, `npos`, is not zero. Better to rework the whole function to use `find_first_of`.

PR #181, opened 2017-07-26
cleaned up internal errors

gitlarryf:


ghewgill:
Despite the user facing keyword being `TRAP`, internally in the compiler it's still called a "catch" (it was called a "catch" even when the user facing keyword was `EXCEPTION`). Notice that the array the loop is iterating through is still called `catches` too. For consistency, and since this error should never actually occur anyway, I recommend leaving the internal_error message the way it is.

PR #187, opened 2018-02-26
C executor

gitlarryf:
For review.

### DIFF
--- a/contrib/NeonVM/cell.c
+++ b/contrib/NeonVM/cell.c
@@ -20,7 +20,7 @@ Cell *cell_fromString(const char *s)
 {
     Cell *x = cell_newCell();
     x->type = String;
-    x->string = _strdup(s);
+    x->string = strdup(s);
     return x;
 }
 
@@ -35,7 +35,7 @@ Cell *cell_fromCell(const Cell *c)
             cell_copyCell(x, c);
             break;
         case String:
-            x->string = _strdup(c->string);
+            x->string = strdup(c->string);
             x->address = NULL;
             x->number = bid128_from_uint32(0);
             break;
@@ -67,7 +67,7 @@ void cell_copyCell(Cell *dest, const Cell *source)
     assert(dest != NULL);
 
     dest->number = source->number;
-    dest->string = _strdup(source->string);
+    dest->string = strdup(source->string);
     dest->address = source->address;
     dest->type = source->type;
 }
--- a/contrib/NeonVM/cell.c
+++ b/contrib/NeonVM/cell.c
@@ -1,6 +1,6 @@
 #include <assert.h>
 #include <stdarg.h>
-#include <stdbool.h>
+//#include <stdbool.h>
 #include <stdint.h>
 #include <stdio.h>
 #include <stdlib.h>
@@ -11,18 +11,18 @@
 
 Cell *cell_fromAddress(Cell *c)
 {
-    Cell *x = cell_newCell();
-    x->type = Address;
+    Cell *x = cell_newCell(cNothing);
+    x->type = cAddress;
     x->address = c;
     return x;
 }
 
 Cell * cell_fromArray(Cell * c)
 {
-    uint32_t i;
+    uint64_t i;
     Cell *a = cell_newCell();
 
-    a->type = Array;
+    a->type = cArray;
     a->array_size = c->array_size;
     a->array = malloc(sizeof(Cell) * c->array_size);
     for (i = 0; i < c->array_size; i++) {
@@ -31,27 +31,38 @@ Cell * cell_fromArray(Cell * c)
     return a;
 }
 
-Cell *cell_fromBoolean(bool b)
+Cell *cell_fromBoolean(BOOL b)
 {
     Cell *x = cell_newCell();
-    x->type = Boolean;
+    x->type = cBoolean;
     x->boolean = b;
     return x;
 }
 
-Cell *cell_fromNumber(BID_UINT128 n)
+Cell *cell_fromNumber(Number n)
 {
     Cell *x = cell_newCell();
-    x->type = Number;
+    x->type = cNumber;
     x->number = n;
     return x;
 }
 
-Cell *cell_fromString(const char *s)
+Cell *cell_fromString(const char *s, int64_t length)
 {
     Cell *x = cell_newCell();
-    x->type = String;
-    x->string = strdup(s);
+    x->type = cString;
+    x->string = string_newString();
+
+    if (length == -1) {
+        x->string->length = strlen(s);
+    } else {
+        x->string->length = length;
+    }
+    x->string->data = malloc(x->string->length);
+    if (!x->string->data) {
+        fatal_error("Failed to allocate memory for new cell string.");
+    }
+    memcpy(x->string->data, s, x->string->length);
     return x;
 }
 
@@ -63,51 +74,60 @@ Cell *cell_fromCell(const Cell *c)
     Cell *x = cell_newCell();
     x->type = c->type;
     switch (c->type) {
-        case Address:
+        case cAddress:
             cell_copyCell(x, c);
             break;
-        case Array:
+        case cArray:
             x->array_size = c->array_size;
             x->array = malloc(sizeof(Cell) * c->array_size);
             for (i = 0; i < x->array_size; i++) {
                 cell_copyCell(&x->array[i], &c->array[i]);
             }
-            x->boolean = false;
+            x->boolean = FALSE;
             x->number = bid128_from_uint32(0);
             x->string = NULL;
+            //x->string->length = 0;
             x->address = NULL;
             break;
-        case Boolean:
+        case cBoolean:
             x->boolean = c->boolean;
             x->number = bid128_from_uint32(0);
             x->string = NULL;
+            //x->string->length = 0;
             x->address = NULL;
             x->array = NULL;
             x->array_size = 0;
             break;
-        case String:
-            x->string = strdup(c->string);
+        case cString:
+            x->string = string_copyString(c->string);
+            //x->string->length = c->string->length;
+            //x->string->data = malloc(x->string->length);
+            //memcpy(x->string->data, c->string->data, x->string->length);
+            //if (!x->string) {
+            //    fatal_error("Failed to allocate memory for cell string.");
+            //}
             x->address = NULL;
             x->number = bid128_from_uint32(0);
-            x->boolean = false;
+            x->boolean = FALSE;
             x->array = NULL;
             x->array_size = 0;
             break;
-        case Number:
+        case cNumber:
             x->number = c->number;
             x->string = NULL;
             x->address = NULL;
-            x->boolean = false;
+            x->boolean = FALSE;
             x->array = NULL;
             x->array_size = 0;
             break;
-        case None:
-            assert(c->type == None);
+        case cNothing:
+            assert(c->type == cNothing);
+            break;
         default:
             x->number = bid128_from_uint32(0);
             x->string = NULL;
             x->address = NULL;
-            x->boolean = false;
+            x->boolean = FALSE;
             x->array = NULL;
             x->array_size = 0;
             break;
@@ -119,7 +139,7 @@ Cell *cell_createArrayCell(uint64_t iElements)
 {
     Cell *c = cell_newCell();
 
-    c->type = Array;
+    c->type = cArray;
     c->array_size = iElements;
     c->array = malloc(sizeof(Cell) * c->array_size);
     if (c->array == NULL) {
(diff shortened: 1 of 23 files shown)


contrib/NeonVM/cell.c
```diff
@@ -0,0 +1,76 @@
...
+    Cell *x = malloc(sizeof(Cell));
+    if (x == NULL) {
+        fatal_error("Could not allocate new number object.");
+    }
+    x->Type = Number;
+    x->number = n;
+    return x;
+}
+
+Cell *cell_fromString(char *s)
```
ghewgill: Use 'const char *' here

contrib/NeonVM/cell.c
```diff
@@ -0,0 +1,76 @@
...
+    Cell *x = malloc(sizeof(Cell));
+    if (x == NULL) {
+        fatal_error("Could not allocate new string object.");
+    }
+    x->Type = String;
+    x->string = s;
+    return x;
+}
+
+Cell *cell_fromCell(Cell *c)
```
ghewgill: Use 'const Cell *' here (in parameter list)

contrib/NeonVM/cell.c
```diff
@@ -0,0 +1,76 @@
...
+        default:
+            x->number = bid128_from_uint32(0);
+            x->string = NULL;
+    }
+    return x;
+}
+
+Cell *cell_fromAddress(Cell *c)
+{
+    Cell *x = cell_newCell();
```
ghewgill: I think you want to fill in x->address before returning x here.

contrib/NeonVM/neonvm.c
```diff
@@ -255,27 +261,87 @@ int main(int argc, char* argv[])
...
     struct tagTExecutor *r = malloc(sizeof(struct tagTExecutor));
     if (r == NULL) {
         fatal_error("Failed to allocate memory for Neon executable.");
     }
     r->object = object;
-    r->callstack = createStack(300);
     r->stack = createStack(300);
     r->ip = 0;
+    r->callstacktop = 0;
+    r->globals = malloc(sizeof(Cell) * r->object->global_size);
```
ghewgill: Looks like this should be sizeof(Cell *) instead of sizeof(Cell), since globals in array of pointers.

contrib/NeonVM/neonvm.c
```diff
@@ -298,20 +364,23 @@ void exec_PUSHB()
...
+    push(self->stack, cell_fromString(self->object->strtable[val]));
 }
 
-void exec_PUSHPG()
+void exec_PUSHPG(struct tagTExecutor *self)
 {
-    fatal_error("not implemented");
+    uint32_t addr = exec_getOperand(self);
+    assert(addr < self->object->global_size);
+    push(self->stack, cell_fromAddress(&self->globals[addr]));
```
ghewgill: Do you want to push the pointer to the cell? Or a pointer to the pointer to the cell?

contrib/NeonVM/neonvm.c
```diff
@@ -344,39 +413,48 @@ void exec_LOADB()
...
-    fatal_error("not implemented");
+    self->ip += 1;
+    Cell *addr = top(self->stack)->address; pop(self->stack);
+    push(self->stack, addr);
 }
 
-void exec_LOADS()
+void exec_LOADS(struct tagTExecutor *self)
 {
+    *self = *self;
```
ghewgill: What exactly is this intended to accomplish? If you're trying to suppress an "uninitialized variable" warning, just 'self = self;' is sufficient. You don't need to copy the whole structure to itself.

contrib/NeonVM/neonvm.c
```diff
@@ -600,8 +678,13 @@ void exec_CALLP(struct tagTExecutor *self)
     const uint8_t *func = self->object->strtable[val];
 
     if (strcmp((const char*)func, "print") == 0) {
-        const char *s = pop(self->stack);
+        const char *s = top(self->stack)->string; pop(self->stack);
```
ghewgill: If your pop() frees memory, then you're about to print memory pointed to by 's' that has already been freed.

contrib/NeonVM/neonvm.c
```diff
@@ -600,8 +678,13 @@ void exec_CALLP(struct tagTExecutor *self)
...
 
     if (strcmp((const char*)func, "print") == 0) {
-        const char *s = pop(self->stack);
+        const char *s = top(self->stack)->string; pop(self->stack);
         printf("%s\n", s);
+    } else if (strcmp((const char*)func, "str") == 0) {
+        BID_UINT128 v = top(self->stack)->number; pop(self->stack);
+        char x[50];
+        bid128_to_string(x, v);
+        push(self->stack, cell_fromString(_strdup(x)));
```
ghewgill: Should the strdup() be done here, or should it be done in cell_fromString?

contrib/NeonVM/neonvm.c
```diff
@@ -697,9 +780,11 @@ void exec_JNASSERT()
     fatal_error("not implemented");
 }
 
-void exec_RESETC()
+void exec_RESETC(struct tagTExecutor *self)
 {
-    fatal_error("not implemented");
+    self->ip += 1;
+    Cell *addr = top(self->stack); pop(self->stack);
+    addr = cell_newCell(addr);
```
ghewgill: Not sure whether this does what you want here. I don't think it does.

contrib/NeonVM/stack.c
```diff
@@ -13,7 +13,10 @@ struct tagTStack *createStack(int capacity)
     }
     stack->capacity = capacity;
     stack->top = -1;
-    stack->data = malloc(stack->capacity * sizeof(void *));
+    stack->data = malloc(stack->capacity * sizeof(Cell));
```
ghewgill: Should this be sizeof(Cell *), since data is an array of pointers?

contrib/NeonVM/neonvm.c
```diff
@@ -416,37 +396,38 @@ void exec_LOADB()
 void exec_LOADN(struct tagTExecutor *self)
 {
     self->ip += 1;
-    Cell *addr = top(self->stack)->address; pop(self->stack);
+    Cell *addr = top(self->stack)->address;
     push(self->stack, addr);
+    pop(self->stack);
```
ghewgill: You can't pop after pushing (you just popped the thing you pushed in the line before). You have to pop the stack *before* pushing the new value.

contrib/NeonVM/cell.c
```diff
@@ -68,12 +53,20 @@ Cell *cell_fromCell(const Cell *c)
...
+    x->address = c;
     return x;
 }
 
+void cell_copyCell(Cell *dest, const Cell *source)
+{
+    assert(source != NULL);
+    assert(dest != NULL);
+
+    memcpy(dest, source, sizeof(Cell));
```
ghewgill: A straight memcpy isn't going to do the right then when you copy strings. Because your cell contains a pointer to allocated memory, after doing a cell_copyCell you're going to have two cells that point to the same memory, which means you're going to have a problem when you go to free one of them. This is actually done correctly in cell_fromCell above.

contrib/NeonVM/cell.c
```diff
@@ -10,54 +10,39 @@
...
+            //memcpy(x, c, sizeof(Cell));
             //x->address = c->address;
             break;
         case String:
             x->string = _strdup(c->string);
-            x->number = bid128_from_uint32(0);
-            x->address = NULL;
             break;
         case Number:
             x->number = c->number;
```
ghewgill: You're no longer setting the other members to NULL here. This may cause a problem if a cell goes from being a string to a number, and keeps the old pointer to the string around. When you go to free that cell, it will call free(c->string) even if the cell is not in fact a string type.

contrib/NeonVM/neonvm.c
```diff
@@ -343,13 +358,13 @@ void exec_PUSHB()
 void exec_PUSHN(struct tagTExecutor *self)
 {
     uint32_t val = exec_getOperand(self);
-    push(self->stack, cell_fromNumber(bid128_from_string(self->object->strtable[val])));
+    push(self->stack, cell_fromNumber(bid128_from_string((char*)self->object->strtable[val])));
```
ghewgill: Why do you need to have this cast here?

contrib/NeonVM/neonvm.c
```diff
@@ -343,13 +358,13 @@ void exec_PUSHB()
...
     uint32_t val = exec_getOperand(self);
-    push(self->stack, cell_fromNumber(bid128_from_string(self->object->strtable[val])));
+    push(self->stack, cell_fromNumber(bid128_from_string((char*)self->object->strtable[val])));
 }
 
 void exec_PUSHS(struct tagTExecutor *self)
 {
     uint32_t val = exec_getOperand(self);
-    push(self->stack, cell_fromString(self->object->strtable[val]));
+    push(self->stack, cell_fromString((const char*)self->object->strtable[val]));
```
ghewgill: Why do you need to have this cast here?

contrib/NeonVM/neonvm.c
```diff
@@ -687,7 +687,7 @@ void exec_IND()
 void exec_CALLP(struct tagTExecutor *self)
 {
     uint32_t val = exec_getOperand(self);
-    const uint8_t *func = self->object->strtable[val];
+    const char *func = self->object->strtable[val];
 
     if (strcmp((const char*)func, "print") == 0) {
```
ghewgill: You can now get rid of these useless casts.

contrib/NeonVM/cell.c
```diff
@@ -67,7 +67,7 @@ void cell_copyCell(Cell *dest, const Cell *source)
     assert(dest != NULL);
 
     dest->number = source->number;
-    dest->string = _strdup(source->string);
+    dest->string = strdup(source->string);
```
ghewgill: This will cause a problem if source->string is NULL (like if the source isn't a String cell at all).

contrib/NeonVM/neonvm.c
```diff
@@ -358,7 +327,7 @@ void exec_PUSHB()
 void exec_PUSHN(struct tagTExecutor *self)
 {
     uint32_t val = exec_getOperand(self);
-    push(self->stack, cell_fromNumber(bid128_from_string((char*)self->object->strtable[val])));
+    push(self->stack, cell_fromNumber(bid128_from_string((char*)self->object->strtable[val])));
```
ghewgill: Unnecessary (char *) cast.

contrib/NeonVM/neonvm.c
```diff
@@ -693,8 +659,8 @@ void exec_CALLP(struct tagTExecutor *self)
         const char *s = top(self->stack)->string;
         printf("%s\n", s);
         pop(self->stack);
-    } else if (strcmp((const char*)func, "str") == 0) {
-        BID_UINT128 v = top(self->stack)->number; pop(self->stack);
+    } else if (strcmp((const char*)func, "str") == 0) {
```
ghewgill: Unnecessary cast.

contrib/NeonVM/number.h
```diff
@@ -16,6 +16,10 @@
 #pragma warning(pop)
 #endif
 
+#ifdef _MSC_VER
```
ghewgill: Why is this is number.h, when the only use of strdup is in cell.c? This block should be moved to cell.c.

contrib/NeonVM/cell.c
```diff
@@ -41,29 +37,47 @@ Cell *cell_fromCell(const Cell *c)
...
         case Address:
             cell_copyCell(x, c);
             break;
+        case Array:
+            x->array_size = c->array_size;
+            x->array = malloc(sizeof(Cell) * c->array_size);
+            for (i = 0; i < x->array_size; i++) {
+                cell_copyCell(&x->array[i], &c->array[i]);
+            }
+            x->boolean = c->boolean;
```
ghewgill: Since this is really an array, copying the boolean doesn't matter much, but maybe x->boolean = false;

contrib/NeonVM/cell.c
```diff
@@ -72,6 +86,9 @@ Cell *cell_fromCell(const Cell *c)
             x->string = NULL;
             x->address = NULL;
             x->boolean = false;
+            x->array = NULL;
+            x->array_size = 0;
+
```
ghewgill: There should be a break; here. It looks wrong without one.

contrib/NeonVM/neonvm.c
```diff
@@ -707,9 +718,50 @@ void exec_CALLP(struct tagTExecutor *self)
...
+        strcpy(s, "[");
+        for (uint32_t x = 0; x < a->array_size; x++) {
+            if (strlen(s) > 1) {
+                strcat(s, ", ");
+            }
+            strcat(s, number_to_string(a->array[x].number));
+        }
+        strcat(s, "]");
+        pop(self->stack);
+        push(self->stack, cell_fromString(s));
```
ghewgill: cell_fromString calls strdup() on the passed string but you're not freeing the one you malloced above.

contrib/NeonVM/neonvm.c
```diff
@@ -707,9 +718,50 @@ void exec_CALLP(struct tagTExecutor *self)
...
+    } else if (strcmp(func, "concat") == 0) {
+        char *b = _strdup(top(self->stack)->string); pop(self->stack);
```
ghewgill: _strdup will fail on non-msvc builds. Use strdup, since you've already got the compatibility #define.

contrib/NeonVM/cell.c
```diff
@@ -125,12 +125,48 @@ Cell *cell_createArrayCell(uint32_t iElements)
     if (c->array == NULL) {
         fatal_error("Unable to allocate memory for array.");
     }
-    for (uint32_t i = 0; i < c->array_size; i++) {
+    for (uint64_t i = 0; i < c->array_size; i++) {
```
ghewgill: Better decide how big you want your arrays to be here. Up in cell_fromArray, you use a uint32_t.

contrib/NeonVM/cell.c
```diff
@@ -125,12 +125,48 @@ Cell *cell_createArrayCell(uint32_t iElements)
...
 }
 
+Cell *cell_arrayIndexForRead(Cell *c, size_t i)
+{
+    if (c->type == None) {
+        c->type = Array;
+    }
+    assert(c->type == Array);
+    if (!c->array) {
+        c->array = malloc(sizeof(Cell) * i);
```
ghewgill: Reading from an array must not expand it.

contrib/NeonVM/cell.c
```diff
@@ -125,12 +125,48 @@ Cell *cell_createArrayCell(uint32_t iElements)
...
+}
+
+Cell *cell_arrayIndexForWrite(Cell *c, size_t i)
+{
+    if (c->type == None) {
+        c->type = Array;
+    }
+    assert(c->type == Array);
+    if (!c->array) {
+        c->array = malloc(sizeof(Cell) * i);
```
ghewgill: Why are you allocating space for i elements here? Even if this path is taken, you're going to realloc next anyway.

contrib/NeonVM/cell.c
```diff
@@ -125,12 +125,48 @@ Cell *cell_createArrayCell(uint32_t iElements)
...
+        if (c->array == NULL) {
+            fatal_error("Unable to allcoate memory for Read Array.");
```
ghewgill: This is writing, not reading.

contrib/NeonVM/cell.c
```diff
@@ -125,12 +125,48 @@ Cell *cell_createArrayCell(uint32_t iElements)
...
+        }
+    }
+
+    if (i >= c->array_size) {
+        c->array = realloc(c->array, i+1);
```
ghewgill: Would you like to initialize these new cells too? I bet you do.

contrib/NeonVM/neonvm.c
```diff
@@ -732,6 +804,86 @@ void exec_CALLP(struct tagTExecutor *self)
         push(self->stack, a);
         cell_freeCell(left);
         cell_freeCell(right);
+    } else if (strcmp(func, "array__size") == 0) {
```
ghewgill: I think it's definitely time to break each of these out into their own functions!

contrib/NeonVM/neonvm.c
```diff
@@ -746,16 +898,26 @@ void exec_CALLP(struct tagTExecutor *self)
...
         strcat(n, b);
         push(self->stack, cell_fromString(n));
         free(a);
         free(b);
         free(n);
+    } else if (strcmp(func, "number__toString") == 0) {
+        BID_UINT128 v = top(self->stack)->number; pop(self->stack);
+        push(self->stack, cell_fromString(number_to_string(v)));
+    } else if (strcmp(func, "string__append") == 0) {
+        char *b = strdup(top(self->stack)->string); pop(self->stack);
```
ghewgill: After doing this strdup, who frees b? I suggest you don't need to strdup at all, because you're only reading from the string.

PR #191, opened 2018-05-03
Cnex dictionary

gitlarryf:


exec/cnex/cnex.c
```diff
@@ -842,19 +848,51 @@ void exec_INDEXAN(TExecutor *self)
...
 
-void exec_INDEXDR()
+void exec_INDEXDR(TExecutor *self)
 {
-    fatal_error("exec_INDEXDR not implemented");
+    self->ip++;
+    TString *index = string_fromString(top(self->stack)->string); pop(self->stack);
+    Cell *addr = top(self->stack)->address; pop(self->stack);
+    Cell *e = dictionary_findDictionaryEntry(addr->dictionary, index);
+    if (!e) {
```
ghewgill: This is a pointer. Use an explicit `== NULL` check.

exec/cnex/cnex.c
```diff
@@ -842,19 +848,51 @@ void exec_INDEXAN(TExecutor *self)
...
 
-void exec_INDEXDW()
+void exec_INDEXDW(TExecutor *self)
 {
-    fatal_error("exec_INDEXDW not implemented");
+    self->ip++;
+    TString *index = string_fromString(top(self->stack)->string); pop(self->stack);
+    Cell *addr = top(self->stack)->address; pop(self->stack);
+    push(self->stack, cell_fromAddress(cell_dictionaryIndexForWrite(addr, index)));
+//    string_freeString(index);
```
ghewgill: Why is this commented out? Who is responsible for freeing the string? It looks like `cell_dictionaryIndexForWrite` doesn't, at least not in all cases.

exec/cnex/cnex.c
```diff
@@ -842,19 +848,51 @@ void exec_INDEXAN(TExecutor *self)
...
 
-void exec_INDEXDV()
+void exec_INDEXDV(TExecutor *self)
 {
-    fatal_error("exec_INDEXDV not implemented");
+    self->ip++;
+    TString *index = peek(self->stack, 0)->string;
+    Cell *dictionary = peek(self->stack, 1);
+    Cell *val = dictionary_findDictionaryEntry(dictionary->dictionary, index);
+    if (!val) {
```
ghewgill: Use explicit `== NULL` test.

exec/cnex/dictionary.c
```diff
@@ -79,12 +79,49 @@ Dictionary *dictionary_createDictionary(int64_t size)
...
+    //    string_clearString(e->key);
+    //}
+
+    return d;
+}
+
+Dictionary *dictionary_copyDictionary(Dictionary *self)
+{
+    Dictionary *d = dictionary_createDictionary(self->len);
+    //d->len = self->len;
```
ghewgill: What's this line for? Seems wrong.

exec/cnex/dictionary.c
```diff
@@ -79,12 +79,49 @@ Dictionary *dictionary_createDictionary(int64_t size)
...
+{
+    if (lhs->len != rhs->len) {
+        return FALSE;
+    }
+
+    Cell *keys = dictionary_getKeys(lhs);
+
+    for (size_t l = 0; l < keys->array_size; l++) {
+        Cell *e = dictionary_findDictionaryEntry(rhs, keys->array[l].string);
+        if (e) {
```
ghewgill: Use explicit check.

exec/cnex/dictionary.h
```diff
@@ -18,12 +18,16 @@ typedef struct tagTDictionary {
...
 void dictionary_freeDictionary(Dictionary *self);
 
-struct tagTCell *dictionary_findDictionaryEntry(Dictionary *self, struct tagTString *key, struct tagTCell *def);
+Dictionary *dictionary_copyDictionary(Dictionary *self);
+int dictionary_compareDictionary(Dictionary *lhs, Dictionary *rhs);
+
+struct tagTCell *dictionary_findDictionaryEntry(Dictionary *self, struct tagTString *key);
 void dictionary_addDictionaryEntry(Dictionary *self, struct tagTString *key, struct tagTCell *value);
 
+int64_t dictionary_findIndex(Dictionary *self, struct tagTString *key);
```
ghewgill: This function is internal and shouldn't be in the header file. Make it `static` in `dictionary.c`.

exec/cnex/dictionary.c
```diff
@@ -96,8 +133,21 @@ void dictionary_freeDictionary(Dictionary *self)
     free(self);
 }
 
+int dictionary_compareKeys(const Cell *a, const Cell *b)
```
ghewgill: Make this helper function `static`.

exec/cnex/number.c
```diff
@@ -117,6 +132,17 @@ Number number_from_sint64(int64_t x)
     return bid128_from_int64(x);
 }
 
+
+/*
+ * Number comparrison functions
```
ghewgill: Spelling.

SConstruct
```diff
@@ -436,6 +436,7 @@ cnex = envcnex.Program("bin/cnex", [
     "exec/cnex/cnex.c",
     "exec/cnex/bytecode.c",
     "exec/cnex/cell.c",
+    "exec/cnex/dictionary.c",
```
ghewgill: Need to change `CMakeLists.txt` too.

exec/cnex/cell.c
```diff
@@ -342,11 +390,12 @@ int32_t cell_compareCell(const Cell * s, const Cell * d)
     }
 
     switch (s->type) {
-        case cString:   return string_compareString(s->string, d->string);
-        case cAddress:  return s->address == d->address != 0;
-        case cBoolean:  return s->boolean == d->boolean != 0;
-        case cArray:    return s->array_size == d->array_size != 0;
-        case cNumber:   return bid128_quiet_equal(s->number, d->number) != 0;
+        case cAddress:      return !(s->address == d->address);
+        case cArray:        return !s->array_size == d->array_size;
```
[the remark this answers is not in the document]
gitlarryf: Fixed.

exec/cnex/cnex.c

exec/cnex/cnex.c

exec/cnex/dictionary.c
gitlarryf: Which line?  The commented out one, or the create_Dictionary() line?

exec/cnex/dictionary.c
gitlarryf: Fixed.

exec/cnex/dictionary.h
gitlarryf: Opps.  Fixed.

exec/cnex/dictionary.c
gitlarryf: Fixed.

exec/cnex/number.c
gitlarryf: D'oh.  Fixed.

SConstruct
gitlarryf: Fixed.

exec/cnex/dictionary.c
ghewgill: The commented one. Why is that there? It would make no sense if it were uncommented, and as a comment it doesn't really help.

gitlarryf: OH.  Right, old code.  Shouldn't be there.

exec/cnex/cnex.c
gitlarryf: The string created by this call get passed to the dictionary object that is created, and becomes owned by the TDictionaryEntry object, and is therefore free()'ed when the Dictionary Object is freed.

exec/cnex/cnex.c
```diff
@@ -47,15 +48,16 @@ void exec_freeExecutor(TExecutor *e)
...
     free(e->globals);
+    assert(isEmpty(e->stack));
     destroyStack(e->stack);
     free(e->stack);
     framestack_destroyFrameStack(e->framestack);
     free(e->framestack);
     free(e->module);
 
-    bytecode_freeBytecode(e->object);
+    //bytecode_freeBytecode(e->object);
```
ghewgill: Why is this commented out? If this function doesn't free the bytecode, then who does?

exec/cnex/cnex.c
```diff
@@ -47,15 +48,16 @@ void exec_freeExecutor(TExecutor *e)
...
 
-    free(e);
+    //free(e);
```
ghewgill: Why is this commented out? If this function doesn't free the executor, then who does? And what should this function be named if it doesn't actually free the executor?

exec/cnex/cnex.c
ghewgill: That's not the case if the dictionary key already exists in the dictionary. The cell_dictionaryIndexForWrite function only holds the index pointer if the element does *not* already exist in the dictionary.

exec/cnex/cnex.c
```diff
@@ -870,9 +911,17 @@ void exec_INA(TExecutor *self)
     push(self->stack, cell_fromBoolean(v));
 }
 
-void exec_IND()
+void exec_IND(TExecutor *self)
 {
-    fatal_error("exec_IND not implemented");
+    self->ip++;
+    Dictionary *dictionary = top(self->stack)->dictionary;
+    TString *key = peek(self->stack, 1)->string;
```
ghewgill: Does this need to be freed in this function? It's unclear to me how this memory is managed.

exec/cnex/dictionary.c
```diff
@@ -40,15 +40,15 @@ static int64_t dictionary_findIndex(Dictionary *self, TString *key)
...
 {
     size_t idx;
 
     idx = dictionary_findIndex(self, key);
-    return idx == -1 ? def : self->data[idx].value;
+    return idx == -1 ? NULL : self->data[idx].value;
 }
 
-void dictionary_addDictionaryEntry(Dictionary *self, TString *key, struct tagTCell *value)
+void dictionary_addDictionaryEntry(Dictionary *self, struct tagTString *key, struct tagTCell *value)
```
ghewgill: As mentioned elsewhere, this function should return the new index.

exec/cnex/dictionary.c
```diff
@@ -59,7 +59,7 @@ void dictionary_addDictionaryEntry(Dictionary *self, TString *key, struct tagTCe
         return;
     }
     if (self->len == self->max) {
-        self->max *= 2;
+        self->max = (self->max * 2) + 1;
```
ghewgill: What is the reason for adding one here?

exec/cnex/dictionary.c
```diff
@@ -79,12 +79,49 @@ Dictionary *dictionary_createDictionary(int64_t size)
     d->len = 0;
     d->max = size;
     d->data = malloc(d->max * sizeof(struct tagTDictionaryEntry));
-    if (!d->data) {
+    if (d->data == NULL) {
         fatal_error("Could not allocate space for %d dictionary entries.", d->max);
     }
+    //for (int64_t i = 0; i < d->len; i++) {
```
ghewgill: What is the purpose of this commented-out code? `d->len` is 0 from a few lines above, so it doesn't make sense here.

exec/cnex/dictionary.c
```diff
@@ -96,8 +133,21 @@ void dictionary_freeDictionary(Dictionary *self)
     free(self);
 }
 
+static  int dictionary_compareKeys(const Cell *a, const Cell *b)
```
ghewgill: To avoid warnings, this should be declared as `qsort()` expects: `(const void *, const void *)`. Then, inside this function, cast the two pointers to `const Cell *`.

exec/cnex/global.c
```diff
@@ -604,9 +631,13 @@ void string__append(TExecutor *exec)
     Cell *addr = top(exec->stack)->address;
 
     addr->string->data = realloc(addr->string->data, addr->string->length + b->string->length);
+    if (addr->string->data == NULL) {
+        fatal_error("Could not allocate %d bytes for string append.",  addr->string->length + b->string->length);
+    }
     memcpy(&addr->string->data[addr->string->length], b->string->data, b->string->length);
     addr->string->length += b->string->length;
 
+    pop(exec->stack);
```
ghewgill: Why pop this at the end? It would make more sense to pop it on line 631.

exec/cnex/number.c
```diff
@@ -131,5 +157,9 @@ void main()
     assert(number_to_string(number_from_string("1234.5678e-3")) == "1.2345678");
     assert(number_to_string(number_from_string("1234.5678e-4")) == "0.12345678");
     assert(number_to_string(number_from_string("1234.5678e-5")) == "0.012345678");
+
+    assert(numer_to_string(number_from_string("123.45")), numer_to_string(number_from_string("123.45")) == TRUE);
```
ghewgill: How does this code even work? There's a spelling error.

exec/cnex/stack.c
```diff
@@ -144,5 +157,14 @@ void callstack_pop(struct tagTCallStack *stack)
 
 size_t callstack_top(struct tagTCallStack *stack)
 {
-    return stack->CallStack.data[stack->CallStack.height];
+    return stack->CallStack.data[stack->CallStack.top];
+}
+
+size_t callstack_peek(struct tagTCallStack *stack, size_t depth)
+{
+    if ((stack->CallStack.top - depth) < 0) {
```
ghewgill: This is a signed/unsigned conflict.

exec/cnex/cell.c
```diff
@@ -342,11 +402,12 @@ int32_t cell_compareCell(const Cell * s, const Cell * d)
...
 
     switch (s->type) {
-        case cString:   return string_compareString(s->string, d->string);
-        case cAddress:  return s->address == d->address != 0;
-        case cBoolean:  return s->boolean == d->boolean != 0;
-        case cArray:    return s->array_size == d->array_size != 0;
-        case cNumber:   return bid128_quiet_equal(s->number, d->number) != 0;
+        case cAddress:      return !(s->address == d->address);
+        case cArray:        return !(s->array_size == d->array_size);
+        case cBoolean:      return !s->boolean == d->boolean;
```
ghewgill: This still isn't right. To be correct, this would have to be `!(s->boolean == d->boolean)`. But instead of using a confusing expression like that, why not use `s->boolean != d->boolean`?

SConstruct
```diff
@@ -433,11 +433,16 @@ else:
     envcnex.Append(CFLAGS=[
         "-std=c99",
```
ghewgill: Should enable warnings here (`-Wall` and `-Werror`).

exec/cnex/cell.c
```diff
@@ -284,11 +308,38 @@ Cell *cell_createDictionaryCell(size_t iEntries)
...
+    if (c->type == cNothing) {
+        c->type = cDictionary;
+    }
+    assert(c->type == cDictionary);
+    if (c->dictionary == NULL) {
+        c->dictionary = dictionary_createDictionary(1);
+    }
+    int64_t idx = dictionary_findIndex(c->dictionary, key);
+    if (idx == -1) {
+        dictionary_addDictionaryEntry(c->dictionary, key, cell_newCell());
```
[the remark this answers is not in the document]
gitlarryf: dictionary_addDictionaryEntry() now returns an int64_t index value.

exec/cnex/dictionary.c
gitlarryf: Removed.  This was an attempt to allocate all of the dictionary records as they were created, but that created a memory management nightmare.

exec/cnex/dictionary.c
gitlarryf: Fixed.

exec/cnex/global.c
gitlarryf: If I did that, then I'd have to create a copy of the object.  Doing it this way, I saved that creation of the object,. and just used it as it existed on the stack, then pop the stack the end to destroy it, and free it.

ghewgill: No, that's not true, because the cell that comes in on the stack is a *pointer* to the real object.

exec/cnex/dictionary.c
```diff
@@ -68,6 +68,7 @@ void dictionary_addDictionaryEntry(Dictionary *self, struct tagTString *key, str
     self->data[self->len].key = key;
     self->data[self->len].value = value;
     self->len++;
+    return self->len;
```
ghewgill: Are you sure this is the right value to return here? Not `self->len - 1`?

exec/cnex/bytecode.c
```diff
@@ -80,7 +80,7 @@ void bytecode_freeBytecode(TBytecode *b)
         string_freeString(b->strings[i]);
     }
     free(b->strings);
-    free(b);
+    //free(b);
```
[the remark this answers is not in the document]
gitlarryf: It is up to the caller free the actual pointer returned by bytecode_newBytecode().  It just seemed to make more sense to do it that way, but please, correct me if I'm wrong, and I can switch this back.

exec/cnex/cnex.c
gitlarryf: It's up to the caller to free the Bytecode chunk individually.  It seemed to make more sense to do it that way when modules got involved.

exec/cnex/cnex.c
gitlarryf: It calls free on the TExecuter that is created, but the caller still has to free the actual pointer that was pointed to by it.  Is that not a good way to handle this?

exec/cnex/cell.c
gitlarryf: Changed to all use the != construct instead.

exec/cnex/bytecode.c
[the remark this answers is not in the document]
ghewgill: Here's the simple test for deciding who does what: (1) Which component calls malloc()? (2) The same component should call free(). The way you've got this currently, the bytecode_newBytecode function calls malloc, and then the caller of *that* is supposed to call free. What if you changed the memory allocation scheme of TBytecode objects? The caller shouldn't have to know or care, all the caller needs to know is a pointer.

exec/cnex/dictionary.c
gitlarryf: I....  I do believe you are right.  After looking at this, yes, I'm incrementing the length, and THEN returning the index, but the index has just been incremented.  So, I"ve now changed this to return the self->len - 1.  This should do the trick.

exec/cnex/cnex.c
```diff
@@ -76,7 +77,9 @@ BOOL ParseOptions(int argc, char* argv[])
             if(argv[nIndex][1] == 'H' || argv[nIndex][1] == 'h' || argv[nIndex][1] == '?' || ((argv[nIndex][1] == '-' && argv[nIndex][2] != '\0') && (argv[nIndex][2] == 'H' || argv[nIndex][2] == 'h'))) {
                 showUsage();
                 exit(1);
-            } else if (argv[nIndex][1] == 'D' || argv[nIndex][1] == 'd') {
+            } else if (argv[nIndex][1] == 'D') {
```
ghewgill: So now some command line options do different things with uppercase and lowercase, and others accept *either* uppercase or lowercase? Be consistent. Choose one style (case-sensitive or not) and stick with it.

exec/cnex/global.c
```diff
@@ -78,6 +78,7 @@ static FILE *check_file(TExecutor *exec, void *pf)
     return f;
 }
 #define DWORD int32_t
+#define GetLastError() 0x02
```
ghewgill: Surely this can't possibly be the right thing to do...

exec/cnex/global.c
```diff
@@ -186,7 +187,7 @@ void file_copy(TExecutor *exec)
 
     //BOOL r = CopyFile(filename, destination, TRUE);
     //if (!r) {
-    //    handle_error(exec, GetLastError(), destination);
+        handle_error(exec, GetLastError(), destination);
```
ghewgill: What is this doing now? Does this function now always fail? Is that by design? Why is the code that does the work commented out?

exec/cnex/util.c
```diff
@@ -28,3 +28,4 @@ void fatal_error(const char *msg, ...)
     va_end(args);
     exit(1);
 }
+
```
ghewgill: What is this extra blank line for? No need to make changes to files that don't have any meaning.

exec/cnex/global.c
gitlarryf: It's sort of a hack to make the compiler happy until go back and implement the file module.

exec/cnex/cnex.c
ghewgill: It's fine to expect command line options that are case sensitive, like `-a` and -A` do different things. It's also fine to expect command line options that are case insensitive, like `-a` and `-A` always do the same thing. But it's not fine to mix *both* styles in the same program. That's the inconsistency.

exec/cnex/global.c
ghewgill: If it does, that would be a serious bug. That's wouldn't be how pointers work in neon bytecode.

exec/cnex/cell.c
```diff
@@ -342,11 +401,14 @@ int32_t cell_compareCell(const Cell * s, const Cell * d)
...
-        case cBoolean:  return s->boolean == d->boolean != 0;
-        case cArray:    return s->array_size == d->array_size != 0;
-        case cNumber:   return bid128_quiet_equal(s->number, d->number) != 0;
+        case cAddress:      return s->address != d->address;
+        case cArray:        return s->array_size != d->array_size;
+        case cBoolean:      return s->boolean != d->boolean;
+        case cDictionary:   return !dictionary_compareDictionary(s->dictionary, d->dictionary);
+        case cNothing:      return s != d;
+        case cNumber:       return !number_is_equal(s->number, d->number);
+        case cPointer:      return (void*)s->address != (void*)d->address;
```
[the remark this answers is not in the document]
gitlarryf: Oh.  I'm not sure that this is exactly what I expected out of this either.  However, without the casts, it would still do what I want it to do.  So, good catch.

exec/cnex/cnex.c
```diff
@@ -75,12 +74,14 @@ BOOL ParseOptions(int argc, char* argv[])
     BOOL Retval = FALSE;
     for (int nIndex = 1; nIndex < argc; nIndex++) {
         if (*argv[nIndex] == '-') {
-            if(argv[nIndex][1] == 'H' || argv[nIndex][1] == 'h' || argv[nIndex][1] == '?' || ((argv[nIndex][1] == '-' && argv[nIndex][2] != '\0') && (argv[nIndex][2] == 'H' || argv[nIndex][2] == 'h'))) {
+            if(argv[nIndex][1] == 'h' || argv[nIndex][1] == '?' || ((argv[nIndex][1] == '-' && argv[nIndex][2] != '\0') && (argv[nIndex][2] == 'h'))) {
```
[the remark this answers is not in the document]
gitlarryf: Yes...  It actually does.  It intended to test for -h, -help, --h, --help, all in one shot.  Since I don't have any --hamburger parameters, it just assumed that you meant "help".

exec/cnex/cnex.c
```diff
@@ -75,12 +74,14 @@ BOOL ParseOptions(int argc, char* argv[])
...
                 exit(1);
-            } else if (argv[nIndex][1] == 'D' || argv[nIndex][1] == 'd') {
+            } else if (argv[nIndex][1] == 'D') {
+                gOptions.ExecutorDisassembly = TRUE;
+            } else if (argv[nIndex][1] == 'd') {
                 gOptions.ExecutorDebugStats = TRUE;
-            } else if (argv[nIndex][1] == 'N' || argv[nIndex][1] == 'n') {
+            } else if (argv[nIndex][1] == 'n') {
                 gOptions.EnableAssertions = FALSE;
             }
```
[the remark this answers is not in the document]
gitlarryf: Done.

exec/cnex/cnex.c
```diff
@@ -10,14 +10,15 @@
 #include <stdarg.h>
 #include <stdint.h>
 #include <stdio.h>
-#include <time.h>
+//#include <time.h>
```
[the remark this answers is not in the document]

exec/cnex/cnex.c
```diff
@@ -268,7 +272,7 @@ void exec_rtl_raiseException(TExecutor *self, const char *name, const char *info
 
     exec_raiseLiteral(self, n, i, code);
     string_freeString(n);
-    string_freeString(i);
+    //string_freeString(i);
```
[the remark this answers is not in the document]

exec/cnex/cnex.c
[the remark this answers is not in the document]
ghewgill: But what if you did want a `--hamburger` parameter? This line would prevent you from adding one. But adding a `--pizza` parameter would work fine. You should be able to add any new parameters without changing your test for the help command.

exec/cnex/cnex.c
```diff
@@ -1009,7 +1063,8 @@ void exec_EXCEPT(TExecutor *self)
     self->ip++;
     unsigned int val = exec_getOperand(self);
     self->ip = start_ip;
-    Cell *exinfo = cell_fromCell(top(self->stack)); pop(self->stack);
+    //Cell *exinfo = cell_fromCell(top(self->stack)); pop(self->stack);
```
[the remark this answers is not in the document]

exec/cnex/cnex.c
```diff
@@ -1020,11 +1075,12 @@ void exec_EXCEPT(TExecutor *self)
     if (size >= 2) {
         code = exinfo->array[1].number;
     }
-    cell_freeCell(exinfo);
+    //cell_freeCell(exinfo);
```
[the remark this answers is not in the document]
gitlarryf: Done.

exec/cnex/cnex.c
```diff
@@ -1051,50 +1107,50 @@ void exec_RESETC(TExecutor *self)
...
+void exec_PUSHCI(void)
 {
     fatal_error("exec_PUSHCI not implemented");
 }
 
 void exec_loop(struct tagTExecutor *self)
 {
     while (self->ip < self->object->codelen) {
-        //if (self->debug) { fprintf(stderr, "mod\t%s\tip:%d\top:\t%s\tst\t%d\n", self->module->name, self->ip, sOpcode[self->object->code[self->ip]], self->stack->top); }
+        if (self->disassemble) { fprintf(stderr, "mod\t%s\tip:%d\top:\t%s\tst\t%d\n", self->module->name, self->ip, sOpcode[self->object->code[self->ip]], self->stack->top); }
```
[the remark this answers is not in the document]
gitlarryf: Fixed.  Also fixed the \t formatting in the fprintf() statement to be more compatible with neonx.exe.

exec/cnex/global.c
```diff
@@ -601,12 +577,16 @@ void number__toString(TExecutor *exec)
...
+    Cell *addr = top(exec->stack)->address; pop(exec->stack);
 
     addr->string->data = realloc(addr->string->data, addr->string->length + b->string->length);
+    if (addr->string->data == NULL) {
+        fatal_error("Could not allocate %d bytes for string append.",  addr->string->length + b->string->length);
+    }
     memcpy(&addr->string->data[addr->string->length], b->string->data, b->string->length);
     addr->string->length += b->string->length;
 
+    //pop(exec->stack);
```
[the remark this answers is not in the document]
gitlarryf: Done.

exec/cnex/stack.c
```diff
@@ -28,6 +28,7 @@ TStack *createStack(int capacity)
 
 void destroyStack(TStack *stack)
 {
+    assert(isEmpty(stack));
     while (!isEmpty(stack)) {
```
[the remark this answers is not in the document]
gitlarryf: That assert wasn't always there.  It was commented out.  Which is better?  The assert, or the loop to clear the stack, which.../should/ already be empty--but might not be?

exec/cnex/stack.c
```diff
@@ -103,37 +108,46 @@ TCallStack *createCallStack(int capacity)
...
+    s->peek = callstack_peek;
     return s;
 }
 
-void destroyCallStack(TCallStack *stack)
+void destroyCallStack(struct tagTCallStack *stack)
 {
     if (stack) {
+        assert(stack->isEmpty(stack));
         if (stack->CallStack.data) {
```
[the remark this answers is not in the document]
gitlarryf: Fixed.

exec/cnex/stack.c
```diff
@@ -103,37 +108,46 @@ TCallStack *createCallStack(int capacity)
...
             free(stack->CallStack.data);
         }
         free(stack);
     }
 }
 
+int callstack_isEmpty(struct tagTCallStack *stack)
+{
+    return stack->CallStack.top == - 1;
```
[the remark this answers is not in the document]
gitlarryf: Opps.  Good catch, once again...

exec/cnex/stack.c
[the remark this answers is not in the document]
ghewgill: The assert is better, because you're saying "if we get here, then the stack should be empty". Otherwise, there's a bug somewhere else. What's the worst that can happen if your program logic is broken and the assertion doesn't hold? A memory leak. If your logic is broken then you've got bigger problems than a memory leak.

exec/cnex/dictionary.c
```diff
@@ -42,7 +42,8 @@ int64_t dictionary_addDictionaryEntry(Dictionary *self, struct tagTString *key,
         return idx;
     }
     if (self->len == self->max) {
-        self->max = (self->max * 2) + 1;
+        //self->max = (self->max * 2) + 1;
```
ghewgill: Remove commented out code.

exec/cnex/dictionary.c
```diff
@@ -62,6 +63,9 @@ Dictionary *dictionary_createDictionary(int64_t size)
     }
     d->len = 0;
     d->max = size;
+    if (size == 0) {
```
ghewgill: Only if it's zero? What if the caller passes 1? Actually, why does the caller even need to pass a `size` here? Does it make sense to do that?

gitlarryf: It did to me, at the time, because of the CONSD opcode, we already know the size of the array.  I've now removed all size parameters on dictionaries.

exec/cnex/number.h
```diff
@@ -5,6 +5,12 @@
 #include <stdio.h>
 #include <stdlib.h>
 
+#ifdef _MSC_VER
+#define strcasecmp _stricmp
+#else
+#define strcasecmp stricmp
```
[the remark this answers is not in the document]
gitlarryf: This was an attempt to get around the MS Compiler warnings, but I see the problem you're pointing out here. -- Fixed.

PR #198, opened 2018-11-02
Add generic Cell toString function

gitlarryf:


exec/cnex/cell.c
```diff
@@ -239,6 +247,71 @@ BOOL cell_arrayElementExists(const Cell *a, const Cell *e)
...
+    assert(c->type != cNothing);
+    TString *r = string_newString();
+
+    switch (c->type) {
+        case cAddress:
+            break;
+        case cArray:
+        {
+            size_t x, y;
+            /* ToDo: Fix this to use proper string appends!! */
```
ghewgill: You've already got append functions for TString. Use those here instead of an unchecked fixed memory allocation.

exec/cnex/cell.c
```diff
@@ -239,6 +247,71 @@ BOOL cell_arrayElementExists(const Cell *a, const Cell *e)
...
+            char *s = malloc(STRING_BUFFER_SIZE);
+            if (s == NULL) {
+                fatal_error("Could not allocate %d byte string buffer for array", STRING_BUFFER_SIZE);
+            }
+            strcpy(s, "[");
+            for (x = 0; x < c->array->size; x++) {
+                if (strlen(s) > 1) {
+                    strcat(s, ", ");
+                }
+                if (c->array->data[x].type == cString) {
```
ghewgill: I'm not sure why this needs to look into the type of each element. This should simply recurse and append the result of cell_toString() for each cell. Right now, you wouldn't be able to handle an array of arrays, for example.

exec/cnex/cell.c
```diff
@@ -239,6 +247,71 @@ BOOL cell_arrayElementExists(const Cell *a, const Cell *e)
...
+            }
+            strcat(s, "]");
+            r = string_appendCString(r, s);
+            free(s);
+            break;
+        }
+        case cBoolean:
+            r = string_appendCString(r, (c->boolean ? "TRUE" : "FALSE"));
+            break;
+        case cNothing:
```
ghewgill: You have already asserted at the top of this function that the type is not cNothing.

exec/cnex/cnex.c
```diff
@@ -120,7 +120,7 @@ int main(int argc, char* argv[])
     /* ToDo: Remove this!  This is only for debugging. */
     /* gOptions.ExecutorDebugStats = TRUE; */
     _CrtSetDbgFlag(_CRTDBG_ALLOC_MEM_DF | _CRTDBG_LEAK_CHECK_DF);
-    _CrtSetBreakAlloc(218);
+    _CrtSetBreakAlloc(200);
```
ghewgill: Can you remove this from the checked-in code? Your debug code really doesn't need to be here. Just maintain it locally.

gitlarryf: I don't know a good clean way to do that.  So, I'm going to create, yet another pull, but this one wil be the final separation of the code files.  It will include an exec.c, and cnex,.c will just become the main ,c file that contains int main(), and the code for the command line options, etc.  I've been meaning to do this for a while anyway.

exec/cnex/cell.c
gitlarryf: Removed this case.

exec/cnex/cell.c
gitlarryf: Removed all C string functions from this case, and use only string_* functions.

exec/cnex/global.c
```diff
@@ -368,50 +367,18 @@ void array__toBytes__number(TExecutor *exec)
...
-        if (strlen(s) > 1) {
-            strcat(s, ", ");
-        }
-        strcat(s, number_to_string(a->array->data[x].number));
-    }
-    strcat(s, "]");
-    pop(exec->stack);
-    push(exec->stack, cell_fromCString(s));
-    free(s);
+//    Cell *a = cell_fromCell(top(exec->stack)); pop(exec->stack);
```
ghewgill: Why is this code commented out? If it doesn't need to be here, it should be removed.

exec/cnex/global.c
```diff
@@ -368,50 +367,18 @@ void array__toBytes__number(TExecutor *exec)
...
-        for (y = 0; y < a->array->data[x].string->length; y++) {
-            strcat(s, &a->array->data[x].string->data[y]);
-        }
-        strcat(s, "\"");
-    }
-    strcat(s, "]");
-    pop(exec->stack);
-    push(exec->stack, cell_fromCString(s));
-    free(s);
+    //Cell *a = cell_fromCell(top(exec->stack)); pop(exec->stack);
```
ghewgill: Remove commented out code.

exec/cnex/util.h
```diff
@@ -1,8 +1,9 @@
 #ifndef UTIL_H
 #define UTIL_H
 
-#define estr(x)     #x
-#define ENUM(x)     estr(x)
+#define estr(x)                 #x
+#define ENUM(x)                 estr(x)
+#define STRING_BUFFER_SIZE      120
```
ghewgill: This is unused and should be removed.

exec/cnex/cell.c
gitlarryf: Fixed by only checking for `cString` so quotes can be placed around the string.  All other elements are simply the result of a recursion into `cell_toString()`

PR #202, opened 2019-01-14
Implement stack depth semantics for exception-opstack test

gitlarryf:


exec/cnex/cnex.c
```diff
@@ -237,6 +248,9 @@ void exec_raiseLiteral(TExecutor *self, TString *name, TString *info, Number cod
                 TString *handler = self->object->strings[self->object->exceptions[i].exid];
                 if ((string_compareString(name, handler) == 0) || (name->length > handler->length && string_startsWith(name, handler) && name->data[handler->length] == '.')) {
                     self->ip = self->object->exceptions[i].handler;
+                    while (self->stack->top > ((framestack_isEmpty(self->framestack) ? 0 : self->framestack->top) + (int32_t)self->object->exceptions[i].stack_depth)) {
```
ghewgill: This doesn't look right. What is `self->framestack->top`? I thought the member you added was called `stack_depth`. Consider changing the name of `top` to reflect what it really is the top of (I don't know what that is).

exec/cnex/framestack.c
```diff
@@ -18,6 +18,7 @@ TFrameStack *framestack_createFrameStack(uint32_t capacity)
     frames->capacity = capacity;
     frames->top = -1;
     frames->max = -1;
+    frames->stack_depth = -1;
```
ghewgill: I don't understand what this is doing here. The `stack_depth` element should be a member of each frame, not a member of the frame stack itself.

exec/cnex/framestack.c
```diff
@@ -57,6 +58,7 @@ void framestack_pushFrame(TFrameStack *framestack, TFrame *item)
     if (framestack->max < (framestack->top + 1)) {
         framestack->max++;
     }
+    framestack->stack_depth += item->stack_depth;
```
ghewgill: Why are you adding this here? This doesn't make sense.

exec/cnex/framestack.h
```diff
@@ -5,12 +5,15 @@
 typedef struct tagTFrame {
     struct tagTCell *locals;
     uint32_t frame_size;
+
```
ghewgill: Unnecessary blank line.

exec/cnex/framestack.h
```diff
@@ -5,12 +5,15 @@
...
+    int stack_depth;
 } TFrame;
 
 typedef struct tagTFrameStack {
     int top;
     int capacity;
     int max;
+    int stack_depth;
```
ghewgill: Per the above comments, this shouldn't be here.

exec/cnex/framestack.c
gitlarryf: Fixed.  Removed from the FrameStack structure.

exec/cnex/framestack.h
gitlarryf: Removed.

exec/cnex/framestack.h
gitlarryf: Removed.

exec/cnex/cnex.c
gitlarryf: Removed ambiguity with stack_depth member.  `top` points to the top most item on the stack.

ghewgill: This still isn't right. You're using the depth of the framestack (`self->framestack->top`) instead of the `stack_depth` of the frame on top of the framestack.

gitlarryf: Fixed this to look at the proper stack frame.

exec/cnex/cnex.c
```diff
@@ -70,6 +70,7 @@ BOOL ParseOptions(int argc, char* argv[])
                 exit(1);
             } else if (argv[nIndex][1] == 'D') {
                 gOptions.ExecutorDisassembly = TRUE;
+                gOptions.ExecutorDebugStats = TRUE;
```
ghewgill: Why is `ExecutorDebugStats` set to true here? Isn't that already covered by the `-d` switch? What if you only want to set `ExecutorDisassembly`, there's no longer a way to do it.

PR #203, opened 2019-01-28
Implement all intrinsic functions in cnex

gitlarryf:
I've also enabled all tests that now pass with the new additions in this pull.

### DIFF
--- a/exec/cnex/global.c
+++ b/exec/cnex/global.c
@@ -17,14 +17,22 @@
 #include "util.h"
 
 
+
 #define PDFUNC(name, func)      { name, (void (*)(TExecutor *))(func) }
 
 TDispatch gfuncDispatch[] = {
-    PDFUNC("concat",                    concat),
-    PDFUNC("concatBytes",               concatBytes),
-    PDFUNC("print",                     print),
-    PDFUNC("str",                       str),
-    PDFUNC("ord",                       ord),
+    PDFUNC("chr",                       neon_chr),
+    PDFUNC("concat",                    neon_concat),
+    PDFUNC("concatBytes",               neon_concatBytes),
+    PDFUNC("int",                       neon_int),
+    PDFUNC("max",                       neon_max),
+    PDFUNC("min",                       neon_min),
+    PDFUNC("num",                       neon_num),
+    PDFUNC("odd",                       neon_odd),
+    PDFUNC("ord",                       neon_ord),
+    PDFUNC("print",                     neon_print),
+    PDFUNC("round",                     neon_round),
+    PDFUNC("str",                       neon_str),
 
     PDFUNC("io$fprint",                 io_fprint),
 
@@ -88,7 +96,127 @@ void global_callFunction(const char *pszFunc, struct tagTExecutor *exec)
     fatal_error("global_callFunction(): \"%s\" - invalid or unsupported predefined function call.", pszFunc);
 }
 
-void print(TExecutor *exec)
+void neon_chr(TExecutor *exec)
+{
+    Number x = top(exec->stack)->number; pop(exec->stack);
+
+    if (!number_is_integer(x)) {
+        exec->rtl_raise(exec, "ValueRangeException", "chr() argument not an integer", BID_ZERO);
+        return;
+    }
+    if (number_is_negative(x) || number_is_greater(x, number_from_uint32(0x10ffff))) {
+        exec->rtl_raise(exec, "ValueRangeException", "chr() argument out of range 0-0x10ffff", BID_ZERO);
+        return;
+    }
+
+    Cell *r = cell_createStringCell(0);
+    // ToDo: Implement UTF8 strings here!!
+    string_appendChar(r->string, (char)number_to_uint32(x));
+
+    push(exec->stack, r);
+}
+
+void neon_concat(TExecutor *exec)
+{
+    Cell *b = peek(exec->stack, 0);
+    Cell *a = peek(exec->stack, 1);
+    Cell *r = cell_createStringCell(b->string->length + a->string->length);
+
+    memcpy(r->string->data, a->string->data, a->string->length);
+    memcpy(&r->string->data[a->string->length], b->string->data, b->string->length);
+
+    pop(exec->stack);
+    pop(exec->stack);
+
+    push(exec->stack, r);
+}
+
+void neon_concatBytes(TExecutor *exec)
+{
+    Cell *b = peek(exec->stack, 0);
+    Cell *a = peek(exec->stack, 1);
+    Cell *r = cell_createStringCell(b->string->length + a->string->length);
+
+    memcpy(r->string->data, a->string->data, a->string->length);
+    memcpy(&r->string->data[a->string->length], b->string->data, b->string->length);
+
+    pop(exec->stack);
+    pop(exec->stack);
+
+    push(exec->stack, r);
+}
+
+void neon_int(TExecutor *exec)
+{
+    Number a = top(exec->stack)->number; pop(exec->stack);
+
+    push(exec->stack, cell_fromNumber(number_trunc(a)));
+}
+
+void neon_max(TExecutor *exec)
+{
+    Number b = top(exec->stack)->number; pop(exec->stack);
+    Number a = top(exec->stack)->number; pop(exec->stack);
+
+    if (number_is_greater(a, b)) {
+        push(exec->stack, cell_fromNumber(a));
+    } else {
+        push(exec->stack, cell_fromNumber(b));
+    }
+}
+
+void neon_min(TExecutor *exec)
+{
+    Number b = top(exec->stack)->number; pop(exec->stack);
+    Number a = top(exec->stack)->number; pop(exec->stack);
+
+    if (number_is_greater(a, b)) {
+        push(exec->stack, cell_fromNumber(b));
+    } else {
+        push(exec->stack, cell_fromNumber(a));
+    }
+}
+
+void neon_num(TExecutor *exec)
+{
+    char *str = string_asCString(top(exec->stack)->string); pop(exec->stack);
+    Number n = number_from_string(str);
+    free(str);
+
+    if (number_is_nan(n)) {
+        exec->rtl_raise(exec, "ValueRangeException", "num() argument not a number", BID_ZERO);
+        return;
+    }
+    push(exec->stack, cell_fromNumber(n));
+}
+
+void neon_odd(TExecutor *exec)
+{
+    Number n = top(exec->stack)->number; pop(exec->stack);
+
+    if (!number_is_integer(n)) {
+        exec->rtl_raise(exec, "ValueRangeException", "odd() requires integer", BID_ZERO);
+        return;
+    }
+
+    push(exec->stack, cell_fromBoolean(number_is_odd(n)));
+}
+
+void neon_ord(TExecutor *exec)
+{
+    Cell *s = top(exec->stack);
+
+    if (s->string->length != 1) {
+        exec->rtl_raise(exec, "ArrayIndexException", "ord() requires string of length 1", BID_ZERO);
+        return;
+    }
+    Number r = bid128_from_uint32((uint32_t)s->string->data[0]);
+
+    pop(exec->stack);
+    push(exec->stack, cell_fromNumber(r));
+}
+
+void neon_print(TExecutor *exec)
 {
     const Cell *s = top(exec->stack);
     fwrite(s->string->data, 1, s->string->length, stdout);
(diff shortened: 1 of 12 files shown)


exec/cnex/nstring.c
```diff
@@ -157,6 +202,17 @@ TString *string_appendString(TString *s, TString *ns)
     return s;
 }
 
+TString *string_fillString(TString *s, char *c, size_t count)
```
ghewgill: This should be just `char c`, there's no reason for it to be a pointer to a character.

exec/cnex/global.c
```diff
@@ -97,6 +107,27 @@ void print(TExecutor *exec)
...
+    }
+    if (number_is_negative(x) || number_is_greater(x, number_from_uint32(0x10ffff))) {
+        exec->rtl_raise(exec, "ValueRangeException", "chr() argument out of range 0-0x10ffff", BID_ZERO);
+        return;
+    }
+
+    Cell *r = cell_createStringCell(0);
+    // ToDo: Implement UTF8 strings here!!
+    char ch = (char)number_to_uint32(x);
+    string_appendChars(r->string, (char*)&ch, 1);
```
ghewgill: No cast needed. `&ch` is already a `char *`, you don't need to cast it again.

exec/cnex/global.c
```diff
@@ -127,6 +158,63 @@ void concatBytes(TExecutor *exec)
...
+
+    push(exec->stack, cell_fromNumber(number_trunc(a)));
+}
+
+void format(struct tagTExecutor *exec)
+{
+    TString *fmt   = string_fromString(top(exec->stack)->string); pop(exec->stack);
+    TString *str   = string_fromString(top(exec->stack)->string); pop(exec->stack);
+
+    Spec *spec = malloc(sizeof(Spec));
```
ghewgill: Why is this done using `malloc`? Seems to be no reason for that. There's a memory leak because of this too.

exec/cnex/nstring.c
```diff
@@ -114,6 +114,28 @@ int string_compareString(TString *lhs, TString *rhs)
...
+        end = s->length;
+    }
+
+    if (start <= 0 || start >= s->length) {
+        start = 0;
+    }
+
+    r->data = realloc(r->data, end - start);
+    r->length = end - start;
+    memcpy(r->data, (void*)&s->data[start], end - start);
```
ghewgill: No need for a `(void *)` cast here.

t/rtl.neon
```diff
@@ -57,3 +57,14 @@ print(str(int(3.9)))
 
 print(str(int(-3.9)))
 %= -3
+
+print(strb(odd(3)))
```
ghewgill: Move this up just above `ord` so that there is some sort of order to this file.

exec/cnex/global.c

exec/cnex/format.c
```diff
@@ -0,0 +1,156 @@
...
+            Number x = number_from_string(s);
+            free(s);
+            string_clearString(r);
+            if (number_is_zero(x)) {
+                r = string_appendChars(r, "0", 1);
+            } else {
+                Number base = number_from_uint32(16);
+                // TODO: handle negative numbers
+                while (!number_is_zero(x)) {
+                    r = string_appendChars(r, (char*)&hexdigit[number_to_uint32(number_modulo(x, base))], 1);
```
ghewgill: No `(char*)` cast needed here. It's already a character pointer.

exec/cnex/format.c
```diff
@@ -0,0 +1,156 @@
...
+    static char hexdigit[] = "0123456789abcdef";
+
+    TString *r = str;
+    switch (spec->type) {
+        case 'd': {
+            char *s = string_asCString(str);
+            Number x = number_from_string(s);
+            free(s);
+            if (number_is_nan(x)) {
+                exec->rtl_raise(exec, "ValueRangeException", TCSTR(str), BID_ZERO);
```
ghewgill: Do you need to `return` after this?

exec/cnex/format.h
```diff
@@ -0,0 +1,23 @@
+#ifndef FORMAT_H
+#define FORMAT_H
+#include <stdint.h>
+
+#include "exec.h"
```
ghewgill: You don't need to include all of `exec.h` here. A forward class declaration would do just fine.

exec/cnex/format.h
```diff
@@ -0,0 +1,23 @@
...
+#include "nstring.h"
```
ghewgill: Same with `nstring.h`, because you only ever use pointers to `TString` in this file.

exec/cnex/global.c
```diff
@@ -97,6 +107,27 @@ void print(TExecutor *exec)
...
+    }
+    if (number_is_negative(x) || number_is_greater(x, number_from_uint32(0x10ffff))) {
+        exec->rtl_raise(exec, "ValueRangeException", "chr() argument out of range 0-0x10ffff", BID_ZERO);
+        return;
+    }
+
+    Cell *r = cell_createStringCell(0);
+    // ToDo: Implement UTF8 strings here!!
+    char ch = (char)number_to_uint32(x);
+    string_appendChars(r->string, &ch, 1);
```
ghewgill: There are a lot of instances where you call `string_appendChars(s, &c, 1)` and only append one character. Rather than this inconvenient construction, suggest making a new `string_appendChar(TString *, char)` function that only appends *one* character. That way you can avoid needing a temporary variable like in this case, and the interface becomes simpler.

exec/cnex/global.c
```diff
@@ -139,14 +226,54 @@ void strb(TExecutor *exec)
...
+    Number len = top(exec->stack)->number; pop(exec->stack);
+    Number start = top(exec->stack)->number; pop(exec->stack);
+    TString *ss   = string_fromString(top(exec->stack)->string); pop(exec->stack);
+
+    assert(number_is_integer(start));
+    assert(number_is_integer(len));
+
+    size_t offset = number_to_uint32(start);
+    if (offset > ss->length) {
+        exec->rtl_raise(exec, "ArrayIndexException", "offset", BID_ZERO);
```
ghewgill: Seems you keep forgetting the `return`.

exec/cnex/global.c
```diff
@@ -139,14 +226,54 @@ void strb(TExecutor *exec)
...
+    }
+
+    size_t length = number_to_uint32(len);
+    if (length +offset > ss->length) {
+        exec->rtl_raise(exec, "ArrayIndexException", "length", BID_ZERO);
```
ghewgill: And here too.

exec/cnex/nstring.c
```diff
@@ -145,6 +179,17 @@ void string_resizeString(TString *s, size_t n)
...
 }
 
+TString *string_reverseString(TString *s, size_t start, size_t end)
+{
+    TString *r = string_createString(end-start);
+    for (size_t i = start, e = end; e > start; e--, i++) {
+        r->data[e] = s->data[i];
+    }
+    s->data = r->data;
+    string_freeString(r);
```
ghewgill: Does this do the right thing here? First you assign the `data` pointer from `r` to `s`, and then you free `r`. Doesn't that also free `r->data`, which is also pointed to by `s->data`? Leaving `s->data` pointing to freed memory?

exec/cnex/nstring.c
```diff
@@ -145,6 +179,17 @@ void string_resizeString(TString *s, size_t n)
     s->length = n;
 }
 
+TString *string_reverseString(TString *s, size_t start, size_t end)
+{
+    TString *r = string_createString(end-start);
+    for (size_t i = start, e = end; e > start; e--, i++) {
```
ghewgill: I'm not convinced this loop isn't off by one. When `e` starts, it points to `end`, which is the last character to reverse, correct? But then the continuation test is `e > start`, so `e` will never actually equal `start` inside the loop. So the `start` character never gets copied.

exec/cnex/global.c
```diff
@@ -127,6 +158,62 @@ void concatBytes(TExecutor *exec)
...
+}
+
+void format(struct tagTExecutor *exec)
+{
+    TString *fmt   = string_fromString(top(exec->stack)->string); pop(exec->stack);
+    TString *str   = string_fromString(top(exec->stack)->string); pop(exec->stack);
+
+    Spec spec = { 0 };
+    if (!parse(fmt, &spec)) { // TODO: utf8 if necessary
+        exec->rtl_raise(exec, "FormatException", TCSTR(fmt), BID_ZERO);
```
ghewgill: Needs a `return`?

exec/cnex/format.c
gitlarryf: Removed cast.

exec/cnex/format.c
gitlarryf: I do.  Added return NULL;.

exec/cnex/format.h
gitlarryf: Removed include.

exec/cnex/format.h
gitlarryf: Removed include.

exec/cnex/global.c
gitlarryf: Added missing returns.

exec/cnex/global.c
gitlarryf: Added return.

exec/cnex/nstring.c
gitlarryf: It surely did not.  Fixed this to use string_copyString().

exec/cnex/global.c
gitlarryf: Changed string_appendChars() to string_appendChar().  and it only takes a char, without a count.

exec/cnex/nstring.c
gitlarryf: There was absolutely an issue with this loop.  It has now been fixed.

exec/cnex/format.c
```diff
@@ -0,0 +1,157 @@
...
+        default:
+            return FALSE;
+    }
+    return TRUE;
+}
+
+static TString *fillstr(Spec *spec, int n)
+{
+    TString *r = string_newString();
+    if (spec->fill < 128) {
```
ghewgill: I know why this `< 128` test is here, because I wrote the code you're cloning, but for *your* code it's completely unclear why you are doing this at all. And you don't even need to at this point.

exec/cnex/format.c
```diff
@@ -0,0 +1,157 @@
...
+            char align = spec->align;
+            if (align == 0) {
+                if (spec->type == 'd' || spec->type == 'x') {
+                    align = '>';
+                } else {
+                    align = '<';
+                }
+            }
+            if (align == '<') {
+                r = string_appendString(r, fillstr(spec, space));
```
ghewgill: Who is responsible for freeing the result of `fillstr()`?

exec/cnex/global.c
```diff
@@ -127,6 +157,64 @@ void concatBytes(TExecutor *exec)
...
+
+    push(exec->stack, cell_fromNumber(number_trunc(a)));
+}
+
+void format(struct tagTExecutor *exec)
+{
+    TString *fmt   = string_fromString(top(exec->stack)->string); pop(exec->stack);
+    TString *str   = string_fromString(top(exec->stack)->string); pop(exec->stack);
+
+    //Spec *spec = malloc(sizeof(Spec));
```
ghewgill: What does this comment mean?

exec/cnex/global.c
```diff
@@ -139,14 +227,57 @@ void strb(TExecutor *exec)
...
+
+    push(exec->stack, cell_fromBoolean(number_is_odd(n)));
+}
+
 void ord(TExecutor *exec)
 {
     Cell *s = top(exec->stack);
 
     if (s->string->length != 1) {
+        pop(exec->stack);
```
ghewgill: As discussed, it is not necessary to clean up the stack in this case. It's harmless to do this, but the pop will be taken care of automatically by the exception handler.

exec/cnex/nstring.c
```diff
@@ -114,6 +114,28 @@ int string_compareString(TString *lhs, TString *rhs)
...
+
+    size_t start = startAt;
+    size_t end = start + count;
+
+    if (end > s->length) {
+        end = s->length;
+    }
+
+    if (start <= 0 || start >= s->length) {
+        start = 0;
```
ghewgill: This seems suspect here. You've already used `start` in the calculation of `end` above, and if you change `start` now, then you're not taking into account how that would change `end`. Also, passing `start` past the end of the string causes the substring to be taken from the *beginning* of the string? A better outcome would be an empty string, consistent with what happens when passing `end` past the end of the string.

exec/cnex/nstring.c
```diff
@@ -145,6 +179,17 @@ void string_resizeString(TString *s, size_t n)
     s->length = n;
 }
 
+TString *string_reverseString(TString *s)
+{
+    TString *r = string_createString(s->length);
+    for (size_t i = 0, e = s->length; e > 0; e--, i++) {
+        r->data[e-1] = s->data[i];
+    }
+    s = string_copyString(r);
```
ghewgill: What is the purpose of copying this string here? Can't you simply `return r;`?

exec/cnex/nstring.c
```diff
@@ -157,6 +202,17 @@ TString *string_appendString(TString *s, TString *ns)
     return s;
 }
 
+TString *string_fillString(TString *s, char c, size_t count)
+{
+    s->data = realloc(s->data, s->length + count);
```
ghewgill: So this `realloc()` seems to indicate that this function is going to *append* the given number of characters, right? That's why you are computing `s->length + count`? But then the implementation just smacks the filled characters over top of the *beginning* of the given string, leaving some of the newly allocated space at the end of the buffer uninitialized. That doesn't seem right.

exec/cnex/format.c
gitlarryf: Removed entire block.

exec/cnex/global.c
gitlarryf: Removed.

exec/cnex/global.c
gitlarryf: Removed unnecessary pop.

exec/cnex/nstring.c
gitlarryf: I'm not sure what you're suggesting I do here.  If start is <= 0, or greater than the length of the string, I should just return a blank string?

exec/cnex/nstring.c
gitlarryf: The function was supposed to modify the string in-place.  I guess it would be better to return a whole different string pointer, not altering the original.

exec/cnex/nstring.c
ghewgill: It seems to me that certainly if start is past the end of the string, you should return an empty string. If the start less than zero, it's up to you. You could return an error, or you could adjust things to return only the part of the (start of the string) that is covered by the start and count.

exec/cnex/nstring.c
ghewgill: Well, that's what it does now, it doesn't modify the passed string in-place.

exec/cnex/nstring.c
ghewgill: Well, what do you *need* it to do? Don't do anything more than what you need at this point.

exec/cnex/global.c
```diff
@@ -88,15 +96,27 @@ void global_callFunction(const char *pszFunc, struct tagTExecutor *exec)
     fatal_error("global_callFunction(): \"%s\" - invalid or unsupported predefined function call.", pszFunc);
 }
 
-void print(TExecutor *exec)
+void neon_chr(struct tagTExecutor *exec)
```
ghewgill: Inconsistency: Using "struct tagTExecutor" here, and just "TExecutor" in other places.

PR #207, opened 2019-02-08
Add random module

gitlarryf:


exec/cnex/CMakeLists.txt
```diff
@@ -12,10 +16,17 @@ add_executable(cnex
     number.c
     stack.c
     util.c
+    CNEX_RTL
```
ghewgill: This needs to be `${CNEX_RTL}` to expand the variable. (I think older versions of cmake would automatically expand it, but now you have to do it explicitly.)

exec/cnex/lib/random.c
```diff
@@ -0,0 +1,14 @@
...
+#include <stdlib.h>
+
+#include "cell.h"
+#include "exec.h"
+#include "number.h"
+#include "stack.h"
+
+void random_uint32(TExecutor *exec)
+{
+    push(exec->stack, cell_fromNumber(number_from_uint32(rand())));
```
ghewgill: The C `rand()` function returns the same sequence of pseudo-random numbers unless it is seeded with something. Because this is usually unexpected behavior, the random number generated should be seeded with a suitably random value before the first call to `rand()`. At some point we can provide an explicit seed operation to return the same sequence of pseudo-random numbers.

exec/cnex/lib/random.h
```diff
@@ -0,0 +1,9 @@
+#ifndef RANDOM_H
+#define RANDOM_H
+#include <stdint.h>
```
ghewgill: This include file does not appear to be needed.

gitlarryf: Removed.

exec/cnex/CMakeLists.txt
gitlarryf: Fixed.

exec/cnex/cnex.c
```diff
@@ -107,6 +108,7 @@ char *getApplicationName(char *arg)
 int main(int argc, char* argv[])
 {
     gOptions.pszExecutableName = getApplicationName(argv[0]);
+    srand((unsigned int)time(NULL));
```
ghewgill: Suggestion: Explicitly call a function like `random_init()` which is implemented in the module code. This one is one line, but others will be more code and you'll want functions.

PR #210, opened 2019-02-21
Add binary module to cnex

gitlarryf:


exec/cnex/lib/binary.c
```diff
@@ -0,0 +1,478 @@
+#include "binary.h"
+
+#include <assert.h>
```
ghewgill: Do you really need all of these includes? Seems like a lot, but I haven't checked them.

exec/cnex/lib/binary.c
```diff
@@ -0,0 +1,478 @@
...
+        return;
+    }
+}
+
+void binary_and32(TExecutor *exec)
+{
+    Number y = top(exec->stack)->number; pop(exec->stack);
+    Number x = top(exec->stack)->number; pop(exec->stack);
+
+    range_check32(exec, x);
```
ghewgill: If `range_check32()` raises an exception, you need to stop what you're doing and exit this function. You can't just proceed and push the result on the stack, because the stack has already been adjusted to match what the exception handler expects (and the top of the stack is supposed to be the exceptioninfo record).

exec/cnex/lib/binary.c
```diff
@@ -0,0 +1,478 @@
...
+    }
+    unsigned int v = number_to_uint32(w);
+    if (b + v > 32) {
+        v = 32 - b;
+    }
+    if (v == 32) {
+        push(exec->stack, cell_fromNumber(x));
+        return;
+    }
+    push(exec->stack, cell_fromNumber(number_from_uint32((number_to_uint32(x) >> b) & (1 << v) - number_to_uint32(BID_ONE))));
```
ghewgill: Don't use `number_to_uint32(BID_ONE)`. That's just doing unnecessary work. Use the literal constant `1`.

exec/cnex/lib/binary.c
```diff
@@ -0,0 +1,478 @@
...
+    Number x = top(exec->stack)->number; pop(exec->stack);
+
+    range_check32(exec, x);
+    range_checkU32(exec, n);
+    unsigned int b = number_to_uint32(n);
+    if (b >= 32) {
+        push(exec->stack, cell_fromBoolean(FALSE));
+        return;
+    }
+    push(exec->stack, cell_fromBoolean((number_to_uint32(x) & number_to_uint32(BID_ONE) << b) != 0));
```
ghewgill: Use parentheses to clarify when using `&` and `<<` in the same expression.

exec/cnex/lib/binary.c
```diff
@@ -0,0 +1,478 @@
...
+    Number x = top(exec->stack)->number; pop(exec->stack);
+
+    range_check64(exec, x);
+    range_checkU64(exec, n);
+    unsigned int b = number_to_uint32(n);
+    if (b >= 64) {
+        push(exec->stack, cell_fromBoolean(FALSE));
+        return;
+    }
+    push(exec->stack, cell_fromBoolean((number_to_uint64(x) & 1ULL << b) != 0));
```
ghewgill: Need parentheses.

exec/cnex/lib/binary.c
```diff
@@ -0,0 +1,478 @@
...
+    if (b >= 64) {
+        push(exec->stack, cell_fromNumber(BID_ZERO));
+        return;
+    }
+    unsigned int v = number_to_uint32(w);
+    if (b + v > 64) {
+        v = 64 - b;
+    }
+    uint64_t z      = number_to_uint64(y);
+    uint64_t mask   = v < 64 ? (number_to_uint64(BID_ONE) << v) - number_to_uint64(BID_ONE) : ~0;
```
ghewgill: More instances of `number_to_uint64(BID_ONE)`. Make sure you get all of these.

gitlarryf: Fixed.  Removed all instances, and replaced with literal 1.

exec/cnex/lib/binary.c
gitlarryf: Fixed.

exec/cnex/lib/binary.c
gitlarryf: Actually, a quick test seems to prove that I don't need any of them.  However, other compilers might feel differently about it?

exec/cnex/lib/binary.c
gitlarryf: Fixed.  range_check()'s now return a BOOL, and exit the module function if an exception has been raised.

exec/cnex/lib/binary.c
gitlarryf: Fixed.

exec/cnex/lib/binary.c
gitlarryf: Fixed.

PR #211, opened 2019-02-22
Fix cnex parameter handling.

gitlarryf:


exec/cnex/cnex.c
```diff
@@ -80,7 +80,11 @@ BOOL ParseOptions(int argc, char* argv[])
             }
         } else {
             Retval = TRUE;
-            gOptions.pszFilename = argv[nIndex];
+            // Once we assign the name of the application we're going to execute, do not continue to look for the name; consider
+            // all others as parameters to the Neon executable.
+            if (!gOptions.pszFilename) {
```
ghewgill: How does this account for command lines like `cnex test.neonx -n` where the `-n` is not an option to `cnex` itself, but an option to `test.neonx`? It looks like this loop continues even after it has found a program to run.

gitlarryf: I realized that after I submitted this pull.  There's two ways that I see this being handled.  The first, is to simply exit the loop as soon as we find a program to execute, even if it isn't an actual .neonx file.  The other is to store up all additional values after we discover the name of the program to run, and them pass them into the sys,neon module as parameters.  I'm thinking that the first option might be the easiest, and therefore the best.

PR #212, opened 2019-02-26
Added sys module to cnex.

gitlarryf:


exec/cnex/global.c
```diff
@@ -114,9 +123,40 @@ static FILE *check_file(TExecutor *exec, void *pf)
     return f;
 }
 
-void global_init()
+void global_init(int argc, char *argv[], int iArgStart)
 {
     random_initModule();
+    VAR_args.address = cell_createArrayCell(0);
+    sys_initModule(argc, argv, iArgStart, VAR_args.address);
+    BuiltinVariables[0].value = VAR_args.address;
```
ghewgill: Isn't `BuiltinVariables[0].value` already initialized to point to `VAR_args`? This assignment seems to be redundant.

exec/cnex/global.c
```diff
@@ -114,9 +123,40 @@ static FILE *check_file(TExecutor *exec, void *pf)
...
+}
+
+void global_shutdown()
+{
+    // ToDo: Figure out which method will actually free the global variables
```
ghewgill: Has this been figured out? If this commented code is not required, remove it.

exec/cnex/global.c
gitlarryf: I thought so to, but strangely enough, if I remove that assignment, the code fails, and `BuildinVariables[0].value` always points to an empty Cell, of type cNone.  I was hoping to get your input on what might be happening here that I'm missing.

exec/cnex/global.c
gitlarryf: Not entirely.  Using the call below to `cell_clearCell()` also `free()`'s the Array in the Cell; but I still end up leaking memory on shutdown, which isn't a huge deal, but it is distracting.  I think I'm going to attempt a Valgind on this code to see if it can tell me more about what is actually happening here.

exec/cnex/lib/sys.h
```diff
@@ -0,0 +1,12 @@
+#ifndef SYS_H
+#define SYS_H
+
+#include "cell.h"
```
ghewgill: Use forward declaration of `Cell` instead of include.

gitlarryf: Fixed.

exec/cnex/global.c
ghewgill: Oh, well I see. `&VAR_args` and `VAR_args.address` are not the same thing.

exec/cnex/global.c
```diff
@@ -114,9 +123,34 @@ static FILE *check_file(TExecutor *exec, void *pf)
...
     random_initModule();
+
+    VAR_args.address = cell_createArrayCell(0);
+    VAR_args.type = cAddress;
+    sys_initModule(argc, argv, iArgStart, VAR_args.address);
+}
+
+void global_shutdown()
+{
+    cell_clearCell(VAR_args.address);
```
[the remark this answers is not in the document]

PR #213, opened 2019-03-01
Fixed +Inf, and other incorrect string to number output

gitlarryf:


exec/cnex/number.c
```diff
@@ -45,7 +45,8 @@ void number_toString(Number x, char *buf, size_t len)
 
     // If we didn't find an exponent marker, then just return the buffer that it is.
     if (*s != 'E') {
-        buf = val;
+        // Number is possibly +Nan, or +Inf.
+        memcpy(buf, val, sizeof(val));
```
ghewgill: There is no guarantee that `sizeof(val)` fits within `len`.

exec/cnex/number.c
```diff
@@ -45,7 +45,8 @@ void number_toString(Number x, char *buf, size_t len)
 
     // If we didn't find an exponent marker, then just return the buffer that it is.
     if (*s != 'E') {
-        buf = val;
+        // Number is possibly +Nan, or +Inf.
+        memcpy(buf, val, len);
```
ghewgill: What happens if `len` is greater than `sizeof(val)`? That would read past the end of the buffer.

PR #217, opened 2019-03-06
Add io module to cnex

gitlarryf:


exec/cnex/lib/io.c
```diff
@@ -0,0 +1,189 @@
...
+    Number count = top(exec->stack)->number; pop(exec->stack);
+    void *pf = top(exec->stack)->address; pop(exec->stack);
+
+    FILE *f = check_file(exec, pf);
+    uint64_t ncount = number_to_uint64(count);
+
+    Cell *r = cell_createStringCell(ncount);
+    size_t n = fread((unsigned char *)r->string->data, 1, ncount, f);
+
+    assert(r->string->length == n);
```
ghewgill: This assert is going to fail the first time you read fewer bytes than you ask for. You'll have to truncate the `r` string to match the number of bytes actually returned by `fread`.

exec/cnex/lib/io.c
```diff
@@ -0,0 +1,189 @@
...
+#include "number.h"
+#include "stack.h"
+#include "nstring.h"
+
+
+static FILE *check_file(TExecutor *exec, void *pf)
+{
+    FILE *f = (FILE *)(pf);
+    if (f == NULL) {
+        exec->rtl_raise(exec, "InvalidFile", "", BID_ZERO);
```
ghewgill: The exception to be raised here is `IoException.InvalidFile`. The same applies to the other places you raise an exception in here.

PR #223, opened 2019-03-30
Add pointer support to cnex

gitlarryf:


exec/cnex/exclude.txt
```diff
@@ -4,7 +4,6 @@ bigint.neon                # assert
 bytes-index.neon           # PUSHPPG
 cal-test.neon              # os.system
 cformat-test.neon          # module
-check-valid.neon           # class
 class-empty.neon           # class
```
ghewgill: This test (`class-empty`) ought to work. There are a couple of others that work now that pointers are in place.

exec/cnex/nstring.c
```diff
@@ -195,6 +195,19 @@ BOOL string_startsWith(TString *self, TString *s)
     return TRUE;
 }
 
+int32_t string_findChar(TString *self, char c)
+{
+    size_t i;
+
+    for (i = 0; i < self->length; i++) {
+        if (self->data[i] == c) {
+            return (int32_t)i;
```
ghewgill: You know how I feel about random casts like this. Why not make this function return a `size_t`?

gitlarryf: I wanted to do this on purpose so that I can return a -1 if something isn't found.  Is there a better way to handle this?  Since size_t is unsigned, I wanted to avoid making it return a size_t.

ghewgill: There's a couple of ways you can do this while returning the index in a `size_t`:

- Return `SIZE_MAX` if something isn't found.
- Return a "found" flag in addition to the position.

There aren't very many elegant ways to handle this in C.

gitlarryf: I agree, fully; which is why I thought this method would be ok.  However, I suppose it could return a SIGNED integer, and then just return SIZE_MAX; since no string is ever going to have the dot out at SIZE_MAN that I should ever have to worry about!

exec/cnex/exclude.txt
gitlarryf: It does.  I'm going to clean up the exclude.txt file with this build.

ghewgill:
Please change the commit message so it's not past tense ("Add pointer support...").

PR #229, opened 2019-05-03
Fix number to string formatter

gitlarryf:
Fix length issues in number to string formatter.

### DIFF
--- a/exec/cnex/number.c
+++ b/exec/cnex/number.c
@@ -65,31 +65,33 @@ void number_toString(Number x, char *buf, size_t len)
     while ((p >= v) && (*p == '0')) p--;
 
     // Once found, we can note the position of it in the string.  Since we know that v is a pointer to the start of the string, and p is the last digit found.
-    const int last_significant_digit = (int)(p - v) + 1;
+    const int last_significant_digit = (int)(p - v);
 
-    if (last_significant_digit == 0) {
+    if (last_significant_digit == -1) {
         buf[0] = '0';
         buf[1] = '\0';
         return;
     }
 
     // Once we know the last siginificant digit, we can calculate how many trailing zero existed.
-    const int trailing_zeros = (int)slen - last_significant_digit;
+    const int trailing_zeros = (int)slen - last_significant_digit-1;
+    // Since we're going to chop of some trailing zeros, we need to update the length of our string.
+    slen -= trailing_zeros;
     exponent += trailing_zeros;
     char mantissa[50] = { 0 };
     char r[50] = { 0 };
     // Copy the mantissa out to the last significant digit.
-    memcpy(mantissa, v, last_significant_digit);
+    memcpy(mantissa, v, last_significant_digit+1);
     // Guarantee the string is null terminated, despite the buffer starting off NULL.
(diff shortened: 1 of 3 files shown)


exec/cnex/number.c
```diff
@@ -116,40 +118,39 @@ void number_toString(Number x, char *buf, size_t len)
             // Then copy the remainder of the mantissa to our [r]esult.
             memcpy(&r[count], mantissa, slen);
             // Ensure that [r]esult is null terminated.  I'm not 100% convinced this is necessary.
-            r[count+slen] = '\0';
+            r[slen += count] = '\0';
```
ghewgill: There's no need to embed this expression inside an array access. Split into two statements.

exec/cnex/number.c
```diff
@@ -510,8 +511,27 @@ BOOL number_is_odd(Number x)
 
 //#define __NUMBER_TESTS
 #ifdef __NUMBER_TESTS
+#include "time.h"
 void main()
 {
+     srand((unsigned int)time(NULL));
```
ghewgill: Misaligned indent

exec/cnex/number.c
```diff
@@ -510,8 +511,27 @@ BOOL number_is_odd(Number x)
...
+
+    for (int i = 0; i < 100000; i++) {
+        BID_UINT128 n = { 0 };
+        n.w[0] = rand();
```
ghewgill: This really isn't going to do what you want here. I'm pretty sure that `rand()` on your platform returns a 32-bit number, but `w[0]` is a 64 bit value. You'll want to make sure that you fill *all* the bits of `n` with random data, not just half of them.

gitlarryf: I checked it, it just returns an int, so under amd64 it works, but I didn't think of making sure it works under x86 also.  Originally, I thought it was 32 bit also, and I tried to do a shift and an or, but that was when I realized it wouldn't let me shift by 32 bits.  I presume I can just do some simple sizeof() checks on what rand() returns, and if it is 4, then do a left sift by 32 bits, and then call rand() again and logically or it with the previous result.  I was also monitoring the numbers I was formatting, and they were some pretty wacky values.  It might be beneficial to have some smaller numbers to format too.

exec/cnex/number.c
gitlarryf: D'oh.

exec/cnex/number.c
ghewgill: I don't think this has anything to do with amd64 vs x86. The import point is, what is `RAND_MAX` on your platform? I would be surprised if it returns a 64 bit random number.

gitlarryf: Fixed with a platform independent method.

exec/cnex/number.c
```diff
@@ -510,8 +512,31 @@ BOOL number_is_odd(Number x)
 
 //#define __NUMBER_TESTS
```
ghewgill: These tests appear to be disabled in the build, and no build file appears to define `__NUMBER_TESTS`. Make sure these tests are enabled and run in the build.

PR #230, opened 2019-05-08
Fix include files

gitlarryf:


exec/cnex/number.c
```diff
@@ -1,14 +1,16 @@
...
 #include <assert.h>
-#include <stdarg.h>
-#include <stdint.h>
 #include <stdio.h>
 #include <stdlib.h>
 #include <string.h>
 
 #include "cell.h"
 
+#ifdef _MSC_VER
```
ghewgill: What is the purpose of this? According to https://docs.microsoft.com/en-us/cpp/c-runtime-library/reference/strcmp-wcscmp-mbscmp?view=vs-2019 there is no `_strcmp`. `strcmp` is a standard function. I don't think this is necessary. Actually, `strcmp` is never called in this module.

gitlarryf: Removed.

PR #231, opened 2019-05-15
Fix LOAD opcodes to ensure cells are of correct type

gitlarryf:


exec/cnex/global.c
```diff
@@ -203,11 +203,11 @@ struct tagTVariables {
 void global_init(int argc, char *argv[], int iArgStart)
 {
     // Init io module
-    VAR_stderr.address = cell_createAddressCell((void*)stderr);
+    VAR_stderr.address = cell_createOtherCell((void*)stderr);
```
ghewgill: You do not need `(void*)` casts here. (It's never needed to cast to `(void*)`, stop doing it, all it does is hide bugs.)

PR #235, opened 2019-06-21
Implement io module using Objects

gitlarryf:


exec/cnex/object.h
```diff
@@ -21,7 +22,9 @@ typedef enum tagNObjectType {
 typedef struct tagTObject {
     ObjectType      type;
     int             refcount;
-    struct tagTCell *cell;
+    void            *cell;
```
ghewgill: If this isn't going to always be a TCell, then it shouldn't be called `cell`.

exec/cnex/lib/io.c
```diff
@@ -38,7 +64,7 @@ static FILE *check_file(TExecutor *exec, void *pf)
 void io_fprint(TExecutor *exec)
```
ghewgill: `io.fprint` no longer exists, it's `io.writeLine` now.

exec/cnex/lib/io.c
```diff
@@ -50,7 +76,7 @@ void io_fprint(TExecutor *exec)
 
 void io_close(TExecutor *exec)
 {
-    void *ppf = peek(exec->stack, 0)->address->other;
+    void *ppf = peek(exec->stack, 0)->object->cell;
 
     FILE *f = check_file(exec, ppf);
     fclose(f);
```
ghewgill: Definitely need to null out the FILE* after closing the file, so you don't try to close it again later.

exec/cnex/lib/io.c
```diff
@@ -65,25 +91,24 @@ void io_open(TExecutor *exec)
 
     const char *m;
     switch (number_to_uint32(mode)) {
-        case ENUM_Mode_read:  m = "r"; break;
-        case ENUM_Mode_write: m = "w+"; break;
+        case ENUM_Mode_read:  m = "rb"; break;
+        case ENUM_Mode_write: m = "wb+"; break;
```
ghewgill: It's `w+b`, not `wb+`.

exec/cnex/object.h
```diff
@@ -13,6 +13,7 @@ typedef enum tagNObjectType {
     oArray,
     oBoolean,
     oBytes,
+    oFile,
```
ghewgill: Probably not needed, as discussed.

gitlarryf: Removed.

exec/cnex/lib/io.c
gitlarryf: Fixed.

exec/cnex/lib/io.c
gitlarryf: Fixed.  Should I also NULL it in release()?

exec/cnex/object.h
gitlarryf: Changed it to `ptr`.  Considered `objPtr`, but decided to go with the shorter name.

exec/cnex/lib/io.c
gitlarryf: To be clear, did you mean `textio.writeLine()`?  There is no exported function writeLine in io.neon/cpp.  I will remove this function from io.c.

ghewgill: Yes, sorry I did mean `textio.writeLine`. Also, there is no `io.readLine`. That's the whole reason for existence of the `textio` module.

exec/cnex/lib/io.c
```diff
@@ -22,6 +22,31 @@
...
+
+void object_releaseFileObject(Object *o)
+{
+    if (o != NULL) {
+        assert(o->refcount > 0);
+        o->refcount--;
+
+        if (o->refcount <= 0) {
+            if (o->ptr != NULL) {
+                fclose(o->ptr);
```
ghewgill: You will probably have to cast this to a `FILE*` for at least some compiler.

exec/cnex/lib/io.c
```diff
@@ -35,25 +60,13 @@ static FILE *check_file(TExecutor *exec, void *pf)
...
-}
-
 void io_close(TExecutor *exec)
 {
-    void *ppf = peek(exec->stack, 0)->address->other;
+    void *ppf = peek(exec->stack, 0)->object->ptr;
 
     FILE *f = check_file(exec, ppf);
     fclose(f);
+    f = NULL;
```
ghewgill: This isn't going to do the trick. You're clearing the local variable `f`, which isn't going to mean much. What you need to clear is the `ptr` member so further file operations can fail cleanly.

exec/cnex/object.c
```diff
@@ -108,8 +109,8 @@ void object_releaseObject(Object *o)
         o->refcount--;
 
         if (o->refcount <= 0) {
-            if (o->cell != NULL) {
-                cell_freeCell(o->cell);
+            if (o->ptr != NULL) {
+                cell_freeCell(o->ptr);
```
ghewgill: This doesn't look right. What if `ptr` isn't a cell? Can that happen here?

exec/cnex/lib/io.c
gitlarryf: Removed io.readLine() also.

exec/cnex/object.c
gitlarryf: No, it shouldn't EVER happen here.  But it is actually the responsibility of the caller to make sure that the default object_releaseObject() never gets called on any non-cell objects.  I have no way to test to see if the object is a Cell or not either.  Unless I add custom enums for the Objects types.

exec/cnex/lib/io.c
gitlarryf: Fixed.

exec/cnex/lib/io.c
gitlarryf: Fixed.

PR #237, opened 2019-07-03
Add textio library to cnex

gitlarryf:


exec/cnex/lib/textio.c
```diff
@@ -0,0 +1,142 @@
...
+#include "nstring.h"
+#include "object.h"
+
+
+static FILE *check_file(TExecutor *exec, void *pf)
+{
+    FILE *f = (FILE *)(pf);
+    if (f == NULL) {
+        exec->rtl_raise(exec, "TextioException.InvalidFile", "", BID_ZERO);
+        // Exception code will return from here, since f is already NULL.
```
ghewgill: You don't seem to *actually* check the return value of `check_file` anywhere it is called, to see whether it returns `NULL`.

gitlarryf: Because it will happen right here, and raise this exception.  Once it does that, the stack will realign itself, and jump back to the caller.  Will it not?  I'm not sure what I'm doing wrong here.

ghewgill: This isn't C++, you don't have exceptions. After you call `rtl_raise`, control will return back here, and then what? You return `NULL` from this function.

gitlarryf: Fixed this in textio and io code, in two separate commits.

PR #242, opened 2019-07-09
Cnex opcode eqa

gitlarryf:


exec/cnex/cell.c
```diff
@@ -596,7 +596,7 @@ int32_t cell_compareCell(const Cell * s, const Cell * d)
 
     switch (s->type) {
         case cAddress:      return s->address != d->address;
-        case cArray:        return s->array->size != d->array->size;
+        case cArray:        return array_compareArray(s->array, d->array);
```
ghewgill: Comparing this to the dictionary case, one of these looks wrong. This case returns the result of `array_compareArray`, while the dictionary case returns the inverse of `dictionary_compareDictionary`. If these aren't wrong, then the functions should be fixed to return consistent values.

exec/cnex/cnex.c
```diff
@@ -754,14 +754,26 @@ void exec_GEY(TExecutor*self)
     push(self->stack, r);
 }
 
-void exec_EQA(void)
+void exec_EQA(TExecutor*self)
```
ghewgill: Why isn't there a space either before or after the `*`? This isn't multiplication.

gitlarryf: Found and fixed several locations where (for some reason) I didn't have a space between these.  Fixed.

exec/cnex/cell.c
gitlarryf: Fixed array compare logic so that it matches the dictionary matching logic.

exec/cnex/array.c
```diff
@@ -58,3 +58,20 @@ Array *array_copyArray(Array *a)
     return r;
 }
 
+int array_compareArray(Array *l, Array *r)
+{
+    if (l->size < r->size) {
```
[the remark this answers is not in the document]
gitlarryf: Changed this check to a not equal condition, which results in a FALSE return value.

PR #246, opened 2019-07-17
Add module support to cnex

gitlarryf:


exec/cnex/bytecode.h
```diff
@@ -15,6 +15,38 @@ typedef struct tagTFunction {
...
+    unsigned char *value;
+} Constant;
+
+typedef struct tagTVariable {
+    unsigned int name;
+    unsigned int type;
+    unsigned int index;
+} Variable;
+
+//type exportfunction struct {
```
ghewgill: What's this? It does not seem to belong here.

exec/cnex/cnex.c
```diff
@@ -28,28 +29,82 @@ typedef struct tagTCommandLineOptions {
...
+    char *p = arg;
+    char *s = p;
+
+    while (*p) {
+        if (*p == '\\' || *p == '/') {
+            s = p;
+        }
+        p++;
+    }
+    return ++s;
```
ghewgill: Using preincrement here is being unnecessarily tricky. It also doesn't work if the arg doesn't contain any slashes at all. Better to just `return s;` here, and use `s = p + 1;` above.

exec/cnex/cnex.c
```diff
@@ -154,45 +196,113 @@ int main(int argc, char* argv[])
...
 
 int exec_run(TExecutor *self, BOOL enable_assert)
 {
-    self->ip = (uint32_t)self->object->codelen;
-    invoke(self, 0);
     self->enable_assert = enable_assert;
+    self->ip = (uint32_t)self->module->bytecode->codelen;
+    invoke(self, self->modules[0], 0);
+
+    for (unsigned int m = 1; m < self->module_count; m++) {
```
ghewgill: I'm not sure this executes the module initialization code in the correct order. It's important to get this right. How is your code different from `exec.cpp` here?

exec/cnex/cnex.c
```diff
@@ -154,45 +196,113 @@ int main(int argc, char* argv[])
...
+    int pass = 0;
+    char filename[256];
+    // For this first case, look in the cwd for the module.
+    sprintf(filename, "%s.neonx", name);
+    FILE *fp = NULL;
+
+load:
+    fp = fopen(filename, "rb");
+    if (fp == NULL) {
+        switch (pass) {
```
ghewgill: This "pass" stuff and the `goto` is confusing to read. Not acceptable. Find a better way.

exec/cnex/stack.h
```diff
@@ -43,4 +43,22 @@ size_t callstack_top(struct tagTCallStack *stack);
 size_t callstack_peek(struct tagTCallStack *stack, int depth);
 int callstack_isEmpty(struct tagTCallStack *stack);
 
+
+typedef struct tagTItemStack {
```
ghewgill: This seems to be declaring a bunch of things that are not defined and never used. Is there a purpose for this "item stack"?

exec/cnex/bytecode.h
gitlarryf: Removed.

exec/cnex/cnex.c
gitlarryf: Changed.

exec/cnex/cnex.c
gitlarryf: It executes the code in the order in which the modules are discovered in code.  I thought that was how `exec.cpp` did it as well?  Perhaps I need to go back and study `exec.cpp` more.

exec/cnex/stack.h
gitlarryf: This was a preview for a stack designed for use with things like storing paths for %neonpath% and .neonpath parsing, etc.  I've decided on a linked list for that stuff anyway.  This was also designed to be used with the GC that I had been mocking up.  This code honestly shouldn't have been committed.

exec/cnex/cnex.c
```diff
@@ -125,15 +167,15 @@ int main(int argc, char* argv[])
     long nSize = ftell(fp);
     fseek(fp, 0, SEEK_SET);
 
-    TBytecode *pModule = bytecode_newBytecode();
-    uint8_t *bytecode = malloc(nSize);
-    if (bytecode == NULL) {
+    TModule *pModule = module_newModule("");
+    pModule->code = malloc(nSize);
```
ghewgill: Why is this done outside the `module_newModule` function? This breaks encapsulation.

exec/cnex/module.c
```diff
@@ -0,0 +1,52 @@
...
+    memset(r, 0x00, sizeof(TModule));
+
+    r->name = malloc(strlen(name) + 1);
+    if (r->name == NULL) {
+        fatal_error("Could not allocate memory for module name: %s", name);
+    }
+    strcpy(r->name, name);
+
+    r->bytecode = bytecode_newBytecode();
+
```
ghewgill: There are a couple of structure members (`code`, `globals`) that are uninitialized here and yet freed in the `module_freeModule` function.

exec/cnex/cnex.c
gitlarryf: I'm not certain that it IS necessary to keep the raw bytes around.  I think that was just part of the design; just clean up everything during shutdown, rather than before.

exec/cnex/cnex.c
gitlarryf: Fixed with the addition of the test case for testing the load order.

exec/cnex/module.c
gitlarryf: Those are allocated elsewhere.  I'm re-working the implementation, however.

gitlarryf: Fixed.

exec/cnex/stack.h
gitlarryf: Removed.

exec/cnex/cnex.c
gitlarryf: Removed.  Fixed with support.c, which encapsulates all of the locating and loading modules.

exec/cnex/cnex.c
gitlarryf: Fixed.  Modules are now initialized in the correct order, the order they were loaded/referenced in.

exec/cnex/cnex.c
ghewgill: This does not appear to have been changed. The original comment still applies.

exec/cnex/cnex.c
```diff
@@ -28,28 +29,82 @@ typedef struct tagTCommandLineOptions {
...
+    char *e = p;
+
+    while (*p) {
+        if (*p == '\\' || *p == '/') {
+            e = p;
+        }
+        p++;
+    }
+
+    if (e == p) {
```
ghewgill: I think this needs some more careful handling here. If `e == p`, what does that mean exactly? I can only see that happening when `*arg == 0`. Why is `e` being decremented by 2? That would put it before `arg`. Then on line 69, you would be allocating 0 bytes, and trying to store things there. Finally, if `e < arg` then `r[e - arg]` on line 76 definitely does not look right.

exec/cnex/cnex.c
ghewgill: I'm not seeing the fix here.

exec/cnex/support.c
```diff
@@ -0,0 +1,212 @@
...
+
+
+static TPath *neon_paths;
+
+static void path_pushPath(const char *source_path)
+{
+    TPath *p = neon_paths;
+
+    TPath *n = malloc(sizeof(TPath));
+    size_t len = strlen(source_path);
```
ghewgill: What's wrong with using `strdup` here?

exec/cnex/support.c
```diff
@@ -0,0 +1,212 @@
...
+        p = p->next;
+    }
+
+    // Add this new path to the paths.
+    p->next = n;
+}
+
+void path_freePaths(TPath *paths)
+{
+    TPath *n = paths->next;
```
ghewgill: Why does this start at the second item in the list? Wouldn't it be simpler to start at the first? Then, along with simplifying the code, it won't segfault if you pass in an empty list.

exec/cnex/support.c
```diff
@@ -0,0 +1,212 @@
...
+        }
+    }
+
+    // Next, check for .neonpath, and process any paths that might exist in it.
+    FILE *fp = fopen(".neonpath", "r");
+    if (fp) {
+        char *path = NULL;
+        size_t len = 0;
+        for (;;) {
+            if (getline(&path, &len, fp) == -1) {
```
ghewgill: It's completely unnecessary to use getline for this. Use `fgets`.

exec/cnex/support.c
```diff
@@ -0,0 +1,212 @@
...
+        }
+        free(path);
+        fclose(fp);
+    }
+}
+ 
+void path_readModule(TModule *m)
+{
+    char modulePath[1024];
+    if (path_existModule(m->name, modulePath) == FALSE) {
```
ghewgill: Testing for the existence of a file before opening it is a race condition (what if the file gets deleted after testing the existence?). This should be refactored to open the file directly instead of calling `access` first.

exec/cnex/support.h
```diff
@@ -0,0 +1,17 @@
+#ifndef SUPPORT_H
+#define SUPPORT_H
+
+#include "util.h"
+
+typedef struct tagTPath {
```
ghewgill: Why is this type declared in the header file? It's not used in the header file anywhere.

exec/cnex/support.c
gitlarryf: Done.

exec/cnex/support.c
gitlarryf: In some cases I can agree that this could be a potential race condition.  However, in the way the code is written, if we believe the file exists, but it is then deleted before we attempt to open it, we will simply throw an error that we were not able to locate that particular module.  I would really like to keep the function path_existModule(), however, instead of returning a BOOL, how about I return a FILE*.  a NULL return indicates that the module does not exist, while a valid FILE pointer will indicate that the file does exist, and that we already have access to it.  Please give me your feedback on this before I attempt to refactor this whole portion.

exec/cnex/support.h
gitlarryf: Initially, the list was stored in cnex.c, but I quickly realized that I didn't need to store it there;  only in support.c.  So it got moved, this got left behind.
Fixed, now.

exec/cnex/support.c
gitlarryf: Refactored; made more simple.  Also added a call *TO* this this function to clean up the memory!

exec/cnex/support.c
ghewgill: That's fine. I never suggested removing the `path_existModule` function (although it should be renamed if you make this change).

exec/cnex/cnex.c
```diff
@@ -269,7 +352,8 @@ void exec_raiseLiteral(TExecutor *self, TString *name, TString *info, Number cod
         if (!framestack_isEmpty(self->framestack)) {
             framestack_popFrame(self->framestack);
         }
-        tip = self->callstack[sp--];
+        tip = self->callstack[sp--].ip;
+        self->module = self->callstack[sp].mod;
```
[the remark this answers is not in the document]
gitlarryf: There should be a test for this.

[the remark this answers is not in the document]

exec/cnex/support.c
gitlarryf: Fixed to now use strdup().

exec/cnex/cnex.c
```diff
@@ -202,18 +251,47 @@ TExecutor *exec_newExecutor(TBytecode *object)
...
+    // Set current modules to module::main.
+    r->module = r->modules[0];
+
+    // Next, iterate the imports, loading all the module code recursively, as necessary.
+    for (unsigned int m = 0; m < r->module->bytecode->importsize; m++) {
+        TModule *mod = module_loadModule(r, r->module->bytecode->strings[r->module->bytecode->imports[m].name]->data, r->module_count);
+        // mod will ALWAYS point to a module.  Either an existing module we've already loaded, or a new
+        // module that has already been pushed onto the module stack.
+        // Allocate the modules globals
+        mod->globals = malloc(sizeof(Cell) * mod->bytecode->global_size);
```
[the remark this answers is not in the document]
gitlarryf: No.  This just means that the TModule pointer has already been setup.  The globals are ONLY allocated here.

exec/cnex/cnex.c
```diff
@@ -28,28 +30,88 @@ typedef struct tagTCommandLineOptions {
...
-        cell_clearCell(&e->globals[i]);
+static char *getApplicationPath(char *arg)
+{
+    BOOL bHavePath = TRUE;
+    char *p = arg;  // p pointer to provided arg.
+    char *e = p;    // e Last path seperator found.
+
+    // Walk the arg string, recording the last position of a path seperator char.
+    while (*p) {
+        if (*p == '\\' || *p == '/' || *p == ':') {
```
ghewgill: Why is this different from the implementation in `getApplicationName`? On some platforms, a colon is a perfectly valid part of a file name, too.

exec/cnex/cnex.c
```diff
@@ -28,28 +30,88 @@ typedef struct tagTCommandLineOptions {
...
+            e = p;
+        }
+        p++;
     }
-    free(e->globals);
+
+    if (e == arg) {
```
ghewgill: This doesn't work properly in the edge case where the last path separator is also the first character, such as `"/foo"`.

exec/cnex/cnex.c
```diff
@@ -28,28 +30,88 @@ typedef struct tagTCommandLineOptions {
...
+        // We didn't find any path seperators.  So Set the return path to simply be the "." current dir.
+        bHavePath = FALSE;
+    }
+
+    // Allocate space for our return value.  Note: We add two extra bytes here for the path.  Since we walked
+    // to the last path seperator, we need one byte for it, plus a byte for our NULL char.
+    char *r = malloc((e - arg) + 2);
```
ghewgill: If you're going to increment `e` anyway in each branch of the following if statement, why not increment it up here, before the malloc, so you don't need to explain why you're adding 2 and not 1 byte.

exec/cnex/cnex.c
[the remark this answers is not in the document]
ghewgill: In that case, the above comment should be changed so it doesn't say "an existing module we've already loaded", if that is not actually the case.

exec/cnex/support.c
```diff
@@ -0,0 +1,181 @@
...
+    char *ne = getenv("NEONPATH");
+    if (ne != NULL) {
+        char *next;
+        char *path = ne;
+        for (;;) {
+            next = strchr(path, PATHSEP);
+            if (next == NULL) {
+                break;
+            }
+            *next = '\0';
```
ghewgill: Haven't you just reimplemented `strtok` here?

exec/cnex/support.c
```diff
@@ -0,0 +1,181 @@
...
+    FILE *fp = fopen(".neonpath", "r");
+    if (fp) {
+        for (;;) {
+            char path[1024];
+            if (fgets(path, sizeof(path), fp) == NULL) {
+                break;
+            }
+            // ToDo: Trim possible leading and trailing spaces (whitespace) off loaded paths.
+            size_t len = strlen(path);
+            if (path[len-1] == '\n') {
```
ghewgill: This doesn't do what the comment above says it does. This only trims a single trailing newline, while the comment says it will trim any leading or trailing whitespace.

exec/cnex/support.c
```diff
@@ -0,0 +1,181 @@
...
+                path[len-1] = '\0';
+            }
+            // Don't bother pushing empty paths.
+            if (len > 1) {
```
ghewgill: This length test seems wrong. You're testing for a length of 2 or more here? So one character path names aren't allowed?

exec/cnex/support.c
gitlarryf: That's left as a ToDo.

exec/cnex/support.c
gitlarryf: I suppose there CAN be single char path names; like... `/`.

exec/cnex/cnex.c

exec/cnex/cnex.c
```diff
@@ -28,28 +39,41 @@ typedef struct tagTCommandLineOptions {
...
-    }
-    free(e->globals);
+    // First, assert that we've emptied the stack during normal execution.
     assert(isEmpty(e->stack));
     destroyStack(e->stack);
     free(e->stack);
+
+    // Next, iterate all allocated modules, freeing them as we go.
+    // Note that we do not free the last (main) module until all others are freed.
+    for (unsigned int m = 1; m < e->module_count; m++) {
```
ghewgill: Perhaps it would be cleaner to run this loop backwards, so that you free module[0] last as part of the loop, rather than having a separate exception just for the main module.

exec/cnex/support.c
```diff
@@ -0,0 +1,222 @@
...
+    char *p = path;
+    char *s = p;
+
+    while (*p) {
+        if (*p == '\\' || *p == '/') {
+            s = p+1;
+        }
+        p++;
+    }
+    if (s[0] == '\\' || s[0] == '/') {
```
ghewgill: Under what conditions could this ever be true? The above loop should skip past all slashes.

exec/cnex/cnex.c

PR #252, opened 2019-08-30
Add time library to cnex

gitlarryf:


exec/cnex/lib/time_darwin.c
```diff
@@ -0,0 +1,29 @@
+#include <stdlib.h>
+
+#include "cell.h"
+#include "exec.h"
+#include "number.h"
+#include "stack.h"
+
+// Note: These must be included after exec/cnex/util.h, which is used by the above includes.
```
ghewgill: I don't understand what this comment means. What does util.h have to do with these system includes? util.h isn't even included above. Why doesn't this work if you move these system includes above the application includes?

exec/cnex/lib/time_posix.c
```diff
@@ -0,0 +1,39 @@
...
+#include "cell.h"
+#include "exec.h"
+#include "number.h"
+#include "stack.h"
+
+
+void time_sleep(TExecutor *exec)
+{
+    // Note: The number on the top of the stack is in SECONDS!
+    uint64_t ms = number_to_uint64(number_multiply(top(exec->stack)->number, number_from_uint32(1000))); pop(exec->stack);
```
ghewgill: This throws away anything less than one millisecond. Why do this? It would be better to preserve the full precision since nanosleep() supports nanosecond resolution.

exec/cnex/lib/time_win32.c
```diff
@@ -0,0 +1,48 @@
+#include <iso646.h>
+#include <stdlib.h>
+
+#include "cell.h"
+#include "exec.h"
+#include "number.h"
+#include "stack.h"
+#include "time.h"
+
+#include <windows.h>
```
ghewgill: Move this above the application includes.

exec/cnex/lib/time_win32.c
```diff
@@ -0,0 +1,48 @@
...
+    LARGE_INTEGER freq;
+    QueryPerformanceFrequency(&freq);
+    PERFORMANCE_FREQUENCY = number_from_uint64(freq.QuadPart);
+}
+
+void time_sleep(TExecutor *exec)
+{
+    Number seconds = top(exec->stack)->number; pop(exec->stack);
+
+    uint32_t ms = number_to_uint32(number_multiply(seconds, number_from_uint32(1000)));
```
ghewgill: What happens if this overflows a uint32_t?

exec/cnex/lib/time_darwin.c
gitlarryf: Because these includes define TRUE and FALSE as 0 and 1.  I declared them part of an enum for added type safety.  So, as long as mine gets declared first, which, it will with those other files included, then these system files will not attempt to redeclare them.

gitlarryf: Fixed.

exec/cnex/lib/time_win32.c
gitlarryf: Fixed.

exec/cnex/lib/time_win32.c
gitlarryf: Nothing?  It would have to sleep for 24 days for that to be an issue.  None of the other time functions have any exception handling around them, so I assumed that since `Sleep()` takes a uint342_t, I'd just declare it as such.

exec/cnex/lib/time_posix.c
gitlarryf: Fixed so that we can actually sleep by nanoseconds now.

exec/cnex/lib/time_posix.c
```diff
@@ -0,0 +1,40 @@
...
+extern Number NANOSECONDS_PER_SECOND;
+
+void time_sleep(TExecutor *exec)
+{
+    // Note: The number on the top of the stack is in SECONDS!
+    uint64_t ns = number_to_uint64(number_multiply(top(exec->stack)->number, NANOSECONDS_PER_SECOND)); pop(exec->stack);
+
+    // So we must get the correct NANOSECOND value to sleep the correct amount of time.
+    struct timespec ts;
+    if (ns > number_to_uint64(NANOSECONDS_PER_SECOND)) {
```
ghewgill: Can you simply use a constant value here instead of converting from a constant to a Number and then back again?

exec/cnex/lib/time_posix.c
```diff
@@ -0,0 +1,40 @@
...
+        ts.tv_sec = (int)(ns / number_to_uint64(NANOSECONDS_PER_SECOND));
+        ts.tv_nsec = (ns - ((long)ts.tv_sec * number_to_uint64(NANOSECONDS_PER_SECOND)));
+   } else {
+        ts.tv_sec = 0;
```
ghewgill: I don't see why this has to be an if/else situation. The calculation above is fine, this else clause is unnecessary.

exec/cnex/lib/time_posix.c
gitlarryf: Yes.  That was my original idea, but since I already had that other Number defined, I tried to be smooth and use that, but it broke the OSX builds.   So...

Fixed.

exec/cnex/lib/time_posix.c
gitlarryf: Removed unnecessary if/else clause.

PR #254, opened 2019-09-23
Add Neon extension support to cnex

gitlarryf:


exec/cnex/cnex.c
```diff
@@ -19,16 +19,30 @@
...
 
+#ifdef _WIN32
+#define PATHSEP "\\"
+#define LIBRARY_NAME_PREFIX ""
+#else
+#define PATHSEP "/"
+#define LIBRARY_NAME_PREFIX "lib"
+#endif
+
+TExtensionModule *g_ExtenstionModules;
```
ghewgill: Spelling ("Extenstion").

exec/cnex/cnex.c
```diff
@@ -19,16 +19,30 @@
...
+extern struct Ne_MethodTable ExtensionMethodTable;
```
ghewgill: You declare this `extern` here, but then provide a definition in this module later. Which is it?

exec/cnex/cnex.c
```diff
@@ -325,6 +344,41 @@ void exec_rtl_raiseException(TExecutor *self, const char *name, const char *info
     string_freeString(n);
 }
 
+struct Ne_MethodTable ExtensionMethodTable = {
```
ghewgill: There's no reason to define this table in *this* source module. If you put it in `extension.c`, then you don't have to list all these functions in `extension.h`.

exec/cnex/cnex.c
```diff
@@ -1310,9 +1364,54 @@ void exec_JUMPTBL(TExecutor *self)
...
+void exec_CALLX(TExecutor *self)
 {
-    fatal_error("exec_CALLX not implemented");
+    self->ip++;
+    uint32_t mod = exec_getOperand(self);
+    uint32_t name = exec_getOperand(self);
+    uint32_t out_param_count = exec_getOperand(self);
+
+    char modname[256];
+    snprintf(modname, 256, "%s/%sneon_%s", path_getPathOnly(self->module->source_path), LIBRARY_NAME_PREFIX, self->module->bytecode->strings[mod]->data);
```
ghewgill: All this seems like a lot of overhead to format up this shared library name on *every* CALLX opcode. Didn't we talk about having an array indexed by the `mod` index?

exec/cnex/cnex.c
```diff
@@ -1310,9 +1364,54 @@ void exec_JUMPTBL(TExecutor *self)
...
+        if (init == NULL) {
+            fatal_error("neon_exec: function Ne_INIT not found in %s\n", modname);
+        }
+        (init)(&ExtensionMethodTable);
+    }
+
+    void_function_t p;
+    char funcname[256];
+    snprintf(funcname, 256, "%s%s", "Ne_", self->module->bytecode->strings[name]->data);
+    p = rtl_foreign_function(self, modname, funcname);
```
ghewgill: There should at least be a TODO comment here to make this a bit more efficient. You don't actually have to call `rtl_foreign_function` on every run of this opcode.

exec/cnex/extension.c
```diff
@@ -0,0 +1,237 @@
+#include "array.h"
```
ghewgill: Always include a module's own header file as the very first include.

exec/cnex/extension.c
```diff
@@ -0,0 +1,237 @@
...
+
+const struct Ne_Cell *parameterlist_get_cell(const struct Ne_ParameterList *list, int i)
+{
+    // Note: The array elements are stored in reverse order, so compute the index from the end.
+    return (struct Ne_Cell*)&((Cell*)list)->array->data[(((Cell*)list)->array->size - i) - 1];
+}
+
+struct Ne_Cell *parameterlist_set_cell(struct Ne_ParameterList *list, int i)
+{
+    return (struct Ne_Cell*)cell_arrayIndexForWrite((Cell*)list, i);
```
ghewgill: The parameters are stored in the same order as for `parameterlist_get_cell`, so you have to do the same subtraction.

exec/cnex/extension.h
```diff
@@ -0,0 +1,39 @@
+#ifndef EXTENSION_H
+#define EXTENSION_H
+
+struct Ne_ParameterList *parameterlist_alloc(int n);
```
ghewgill: As mentioned elsewhere, there's no reason for all these functions to appear in this header file. All you need here is an `extern` declaration for the method table.

exec/cnex/module.c
```diff
@@ -1,16 +1,30 @@
...
+#include <io.h>
+#define strdup _strdup
+#else
+#include <unistd.h>
+#endif
+
 #include "bytecode.h"
 #include "cell.h"
 #include "exec.h"
+#include "module.h"
```
ghewgill: Why did this move? The include for module.h should be first in this file.

ghewgill: Why did this move? The include for module.h should be first in this file.

exec/cnex/module.c
```diff
@@ -1,16 +1,30 @@
...
 #include "nstring.h"
 #include "support.h"
 
 
+
```
ghewgill: Why adding extra blank line here? There were already two, you surely don't need three.

exec/cnex/module.c
ghewgill: Why did this move? The include for module.h should be first in this file.

exec/cnex/rtl_platform.h
```diff
@@ -0,0 +1,7 @@
+#include "number.h"
```
ghewgill: Should have an include guard.

exec/cnex/cnex.c

exec/cnex/cnex.c
gitlarryf: Removed extern decl.

exec/cnex/cnex.c
gitlarryf: We did, but this was sort of a "make it work quick" fix.  I can go back and move these to the TModule struct, if necessary.

exec/cnex/rtl_posix.c
```diff
@@ -0,0 +1,48 @@
+#include "rtl_platform.h"
+
+#include "exec.h"
+#include "support.h"
+
+#include <dlfcn.h>
+#include <map>
```
ghewgill: This is not a C header file.

exec/cnex/cnex.c
gitlarryf: Moved to extension.c, and eliminated extension.h.

exec/cnex/cnex.c
gitlarryf: Added comment for ToDo.  This was something I had totally forgot about adding.

exec/cnex/module.c
gitlarryf: Opps.  I'm not sure why I changed that.   Fixed.

exec/cnex/module.c
gitlarryf: Fixed.

exec/cnex/rtl_platform.h
gitlarryf: Added.

exec/cnex/rtl_posix.c
gitlarryf: Removed.

PR #255, opened 2019-10-02
Add string module to cnex

gitlarryf:


exec/cnex/global.c
```diff
@@ -130,10 +131,17 @@ TDispatch gfuncDispatch[] = {
...
+    PDFUNC("string$lower",              string_lower),
+    PDFUNC("string$split",              string_split),
+    PDFUNC("string$splitLines",         string_splitLines),
+    PDFUNC("string$trimCharacters",     string_trimCharacters),
+    PDFUNC("string$upper",              string_upper),
+
+    // sys - System level calls
     PDFUNC("sys$exit",                  sys_exit),
 
-    // Textio - Text file modules
```
ghewgill: What happened to this line? You've got heading comments for all the other sections.

exec/cnex/lib/string.c
```diff
@@ -0,0 +1,143 @@
...
+void string_split(TExecutor *exec)
+{
+    TString *d = peek(exec->stack, 0)->string; // Delimeter char or string
+    TString *s = peek(exec->stack, 1)->string; // Original String to split
+
+    Cell *r = cell_createArrayCell(0);
+    Cell *e = cell_newCellType(cString); // Temp element container for found strings.
+    size_t i = 0;
+    while (i < s->length) {
+        e->type = cString; // Ensure that we have the correct Cell type to operate on.
```
ghewgill: Why is this necessary? Didn't you just create a new cell of type string on line 48?

exec/cnex/lib/string.c
```diff
@@ -0,0 +1,143 @@
...
+    push(exec->stack, r);
+}
+
+void string_split(TExecutor *exec)
+{
+    TString *d = peek(exec->stack, 0)->string; // Delimeter char or string
+    TString *s = peek(exec->stack, 1)->string; // Original String to split
+
+    Cell *r = cell_createArrayCell(0);
+    Cell *e = cell_newCellType(cString); // Temp element container for found strings.
```
ghewgill: You don't need to actually *allocate* a new cell here, do you? Can't it be a local variable?

exec/cnex/lib/string.c
```diff
@@ -0,0 +1,143 @@
...
+    size_t i = 0;
+    while (i < s->length) {
+        e->type = cString; // Ensure that we have the correct Cell type to operate on.
+        int64_t nd = string_findString(s, i, d);
+        if (nd == -1) {
+            e->string = string_subString(s, i, s->length - i);
+            cell_arrayAppendElement(r, *e);
+            cell_clearCell(e); // Since this calls free() on the string member, we're free to create a new string in this cell.
```
ghewgill: Oh I see, you're trying to reuse the storage for this temporary thing. Too tricky, you don't need to do that, and it makes your code harder to read.

exec/cnex/lib/string.c
```diff
@@ -0,0 +1,143 @@
...
+
+    push(exec->stack, r);
+}
+
+void string_splitLines(TExecutor *exec)
+{
+    TString *s = peek(exec->stack, 0)->string; // Original String to split
+
+    Cell *r = cell_createArrayCell(0);
+    Cell *e = cell_newCellType(cString); // Temp element container for found strings.
```
ghewgill: Same variable reuse pattern, same comments as above.

exec/cnex/lib/string.c
```diff
@@ -0,0 +1,143 @@
...
+        pop(exec->stack);
+        pop(exec->stack);
+        pop(exec->stack);
+        push(exec->stack, cell_createStringCell(0));
+        return;
+    }
+
+    Cell *r = cell_newCellType(cString);
+    // Note: string_subString() creates a new string; it does not modify the existing string in-place.
+    // ToDo: Would it be better to modify the string in place?  This should be tested and benchmarked.
```
ghewgill: I don't understand this comment. This function returns a new string, modifying the original string in place doesn't make sense.

exec/cnex/nstring.c
```diff
@@ -27,6 +27,20 @@ TString *string_createString(size_t length)
...
 }
 
+TString *string_createStringFromData(void *data, size_t len)
+{
+    TString *c = string_newString();
+
+    c->length = len;
+    c->data = malloc(c->length);
+    if (c->data == NULL) {
+        fatal_error("Could not allocate memory (0x%08X bytes) for new data string.", len);
```
ghewgill: Why is this printing the number of bytes in hex? That seems unnecessary. Use decimal.

exec/cnex/nstring.c
```diff
@@ -208,6 +222,111 @@ size_t string_findChar(TString *self, char c)
...
+{
+    for (size_t i = 0; i < self->length; i++) {
+        if (self->data[i] == c) {
+            return FALSE;
+        }
+    }
+    return TRUE;
+}
+
+int64_t string_findFirstNotOf(TString *self, TString *p)
```
ghewgill: Why is this interface different from `string_findFirstOf`? I would have expected them to both take the same types. Should this also have a `pos` parameter?

exec/cnex/nstring.c
```diff
@@ -208,6 +222,111 @@ size_t string_findChar(TString *self, char c)
...
+    }
+
+    if (self == NULL || self->data == NULL || self->length == 0 || pos > self->length) {
+        return r;
+    }
+
+    for (size_t l = pos; l < self->length; l++) {
+        if (self->data[l] == p->data[0]) {
+            r = l;
+            if (self->length < (l + p->length)) {
```
ghewgill: This check doesn't need to be done here, after finding a first character match. You can do it by calculating the `for` loop bounds for line 298. You only need to search as far as `self->length - p->length`, because any further positions can't possibly hold the target.

exec/cnex/nstring.c
```diff
@@ -208,6 +222,111 @@ size_t string_findChar(TString *self, char c)
...
+
+int64_t string_findCString(TString *self, const char *p)
+{
+    int64_t r = -1;
+
+    TString *psz = string_createCString(p);
+    r = string_findString(self, 0, psz);
+    string_freeString(psz);
+    return r;
+}
```
ghewgill: You don't need this, the only place you're calling it is from unit test code. Remove it.

exec/cnex/lib/string.c
gitlarryf: And that's what the comment was sort of noting.  I wasn't sure if subString() should modify the source string, or create a new one.  I opted to create a new one.  I'll fix the comments so they make more sense.

exec/cnex/nstring.c
gitlarryf: I'd like to leave it in here.  I was using it at some point with some other code, but later took it out.  This piece of code could come in handy later on, not to mention in the unit test code.  I guess at the very least I can throw this in the unit test code file, and just use it from there.

exec/cnex/global.c
gitlarryf: Humph.  That's a new one!  Ok, added that comment back in.

exec/cnex/nstring.c
gitlarryf: Modified to take a pos parameter, but left the match string as a TString object.

exec/cnex/nstring.c
gitlarryf: I don't follow, here.  I have to check the remainder of the string, because the starting char could be anywhere in the string.  When I changed this, it broke my string tests, so I'm trying to figure out what went wrong.  It seems that once we reach the char we're looking for, we break out of the loop.  It seems to make sense, but in practice, it's not working as I'd expect.

gitlarryf: Fixed logic to only search viable area of string.

exec/cnex/nstring.c
gitlarryf: Changed to decimal.  (This was originally hex to help stiff out errors that could have been easily located if this error was thrown.  0xFDFDFDFD for example.

exec/cnex/lib/string.c
gitlarryf: It...CAN.  Since all my `cell_()` functions deal with pointers, I just went with a pointer here too.  I've since moved it to a local variable, which is actually the better way to do this.

exec/cnex/lib/string.c
gitlarryf: Removed / Not necessary anymore.

exec/cnex/lib/string.c
gitlarryf: Removed all this garbage and went to a local variable.

exec/cnex/nstring.c
gitlarryf: Moved this chunk of code to test_string_support.c

exec/cnex/lib/string.c
```diff
@@ -0,0 +1,135 @@
...
+    TString *s = peek(exec->stack, 1)->string; // Original String to split
+
+    Cell *r = cell_createArrayCell(0);
+    size_t i = 0;
+    while (i < s->length) {
+        Cell e = { 0, .type = cString };
+        int64_t nd = string_findString(s, i, d);
+        if (nd == -1) {
+            e.string = string_subString(s, i, s->length - i);
+            cell_arrayAppendElement(r, e);
```
ghewgill: The `cell_arrayAppendElement` function makes a copy of the passed-in cell. So you still have to free the string held by `e`.

exec/cnex/nstring.c
ghewgill: You changed some other function to decimal, which is fine, but this one needs to be changed too.

exec/cnex/nstring.c
```diff
@@ -208,6 +222,99 @@ size_t string_findChar(TString *self, char c)
...
+{
+    for (size_t i = 0; i < self->length; i++) {
+        if (self->data[i] == c) {
+            return FALSE;
+        }
+    }
+    return TRUE;
+}
+
+int64_t string_findFirstNotOf(TString *self, size_t pos, TString *p)
```
ghewgill: This interface is still inconsistent with `string_findFirstOf`. Why does `string_findFirstOf` take a `const char *`, and `string_findFirstNotOf` take a `TString *`? All these functions should take the same types, whatever you decide.

exec/cnex/nstring.c
```diff
@@ -208,6 +222,99 @@ size_t string_findChar(TString *self, char c)
...
+    if (self == NULL || self->data == NULL || self->length == 0) {
+        return r;
+    }
+    if (pos > self->length || p->length > self->length || p->length + pos > self->length) {
+        return r;
+    }
+
+    for (size_t l = pos; l < ((self->length - p->length) + 1); l++) {
+        if (self->data[l] == p->data[0]) {
+            r = l;
```
ghewgill: Why make this assignment to `r` when you don't actually know whether this is the value you want to return yet?

exec/cnex/nstring.c
```diff
@@ -208,6 +222,99 @@ size_t string_findChar(TString *self, char c)
...
+            BOOL bFound = TRUE;
+            for (size_t n = 0; n < p->length; n++) {
```
ghewgill: It's redundant to do this comparison for n = 0 because you already know the first character matches.

exec/cnex/nstring.c
```diff
@@ -208,6 +222,99 @@ size_t string_findChar(TString *self, char c)
...
+        if (isNotOneOf(p, self->data[l-1])) {
+            return l;
+        }
+    }
+    return -1;
+}
+
+int64_t string_findString(TString *self, size_t pos, TString *p)
+{
+    int64_t r = -1;
```
ghewgill: You don't seem to need this `r` variable at all. In the first section of this function, simply `return -1` when you detect an error condition.

exec/cnex/nstring.c
```diff
@@ -208,6 +222,99 @@ size_t string_findChar(TString *self, char c)
...
+            r = l;
+            BOOL bFound = TRUE;
+            for (size_t n = 0; n < p->length; n++) {
+                if (self->data[l+n] != p->data[n]) {
+                    bFound = FALSE;
+                    break;
+                }
+            }
+            if (bFound) {
+                return r;
```
ghewgill: Instead of using `r`, you can `return l` here because `l` hasn't changed.

exec/cnex/nstring.c
```diff
@@ -208,6 +222,99 @@ size_t string_findChar(TString *self, char c)
...
+        return r;
+    }
+    if (self == NULL || self->data == NULL || self->length == 0) {
+        return r;
+    }
+    if (pos > self->length || p->length > self->length || p->length + pos > self->length) {
+        return r;
+    }
+
+    for (size_t l = pos; l < ((self->length - p->length) + 1); l++) {
```
ghewgill: `l` is a really terrible local variable name, because it looks so similar to `1`.

exec/cnex/lib/string.c
gitlarryf: Good catch.  I can't believe I missed that.

exec/cnex/nstring.c
gitlarryf: Fixed interface to both take TString's.

exec/cnex/nstring.c
gitlarryf: Because at that time, `l` is the position that the actual match started at.  Then if we decide that a match was in fact made, I want to return the fist position, not the current position.  I suppose it can be done by subtracting the length of the match string from the current position too.

exec/cnex/nstring.c
gitlarryf: Yea, I knew that.  I guess I was just doing lazy comparison, by re-comparing the first char.  Fixed.

exec/cnex/nstring.c
gitlarryf: Removed r.

exec/cnex/nstring.c
gitlarryf: Removed r, and simply return l now.

exec/cnex/nstring.c
gitlarryf: Fixed.

exec/cnex/nstring.c
gitlarryf: Changed to `i` for index.

exec/cnex/lib/string.c
```diff
@@ -0,0 +1,142 @@
...
+    pop(exec->stack);
+
+    push(exec->stack, r);
+}
+
+void string_splitLines(TExecutor *exec)
+{
+    TString *s = peek(exec->stack, 0)->string; // Original String to split
+
+    TString *eol = string_createCString("\r\n");
```
ghewgill: You don't need to allocate/free a new one of these. Create a local `struct TString`, point the pointer at some constant data, set the length, and then you don't need to free it.

exec/cnex/test_string_support.c
```diff
@@ -0,0 +1,82 @@
...
+#include <string.h>
+
+#include "nstring.h"
+
+
+int64_t string_findCString(TString *self, const char *p)
+{
+    int64_t r = -1;
+
+    TString *psz = string_createCString(p);
```
ghewgill: Also here you don't have to allocate a new string. Use a local variable.

exec/cnex/lib/string.c
gitlarryf: Fixed.  Probably a lot of other places I should be doing this type of stuff in, too.

exec/cnex/test_string_support.c
gitlarryf: Ok, but I have to cast away the constness of p to be able to do it.  That, or change the function prototype.  I don't know which one is more appropriate.

PR #256, opened 2019-10-17
Add bytes_decodeToString function to cnex

gitlarryf:


t/bytes-tostring.neon
```diff
@@ -1,3 +1,8 @@
 LET b: Bytes := HEXBYTES "00 01 23 45 67 89 ab cd ef"
 print(b.toString())
 %= HEXBYTES "00 01 23 45 67 89 ab cd ef"
+
+LET c: Bytes := HEXBYTES "48 65 6C 6C 6F 2C 20 57 6F 72 6C 64 21"
+LET h: String := c.decodeToString()
+print(h)
+%= Hello, World!
```
ghewgill: This test doesn't belong in this file. It belongs in `t/string-bytes.neon`, but there is already a test there that covers `decodeToString`.

gitlarryf: Removed test from test file.

PR #259, opened 2019-11-05
Add file module to cnex

gitlarryf:


exec/cnex/nstring.c
```diff
@@ -221,6 +233,35 @@ BOOL string_startsWith(TString *self, TString *s)
...
+    for (si = 0, i = self->length - s->length; i < self->length; si++, i++) {
+        if (tolower(self->data[i]) != tolower(s->data[si])) {
+            return FALSE;
+        }
+    }
+
+    return TRUE;
+}
+
+BOOL string_endsWith(TString *self, TString *s, BOOL bCaseSensetive)
```
ghewgill: Spelling ("sensitive").

Actually this shouldn't be a bool parameter at all. You've already got a function called `string_endsWithNoCase`, so make this one do a case sensitive compare. The caller can call the other one if they want case insensitivity.

exec/cnex/nstring.c
```diff
@@ -221,6 +233,35 @@ BOOL string_startsWith(TString *self, TString *s)
     return TRUE;
 }
 
+BOOL string_endsWithNoCase(TString *self, TString *s)
+{
+    size_t i, si;
+
+    for (si = 0, i = self->length - s->length; i < self->length; si++, i++) {
```
ghewgill: This will fail with undefined behavior if `s->length` is greater than `self->length`.

exec/cnex/test_string_support.c
```diff
@@ -78,6 +79,20 @@ int main()
...
     TString *r = string_subString(s3, first, last - first);
     assert(string_compareString(r, string_createCString("This is String#3's Data.\t\r\n")) == 0);
+    // Test Find operations
+    assert(string_findCharRev(foobar, '.') == 3);
+    assert(string_findCharRev(s2, ' ') == 19);
+    assert(string_findCharRev(foobar, '/') == NPOS);
+
+    size_t lastsep = string_findCharRev(path, '/');
+    TString *fname = NULL;
+    if (lastsep != NPOS) {
```
ghewgill: This isn't a suitable way to write test code. You *know* what the return value is supposed to be, so check it directly with an assert test. Don't put conditional logic in tests like this.

exec/cnex/lib/file.c
```diff
@@ -0,0 +1,120 @@
...
+    Cell *bytes = top(exec->stack);
+    char *filename = string_asCString(peek(exec->stack, 1)->string);
+
+    FILE *f = fopen(filename, "wb");
+    if (!f) {
+        exec->rtl_raise(exec, "FileException.Open", filename, BID_ZERO);
+        goto bail;
+    }
+
+    if (fwrite(bytes->string->data, sizeof(uint8_t), bytes->string->length, f) <= 0) {
```
ghewgill: This should check that you've actually written the number of bytes you asked for. What if it's an incomplete write?

exec/cnex/lib/file_posix.c
```diff
@@ -0,0 +1,383 @@
...
+
+
+static void handle_error(TExecutor *exec, int error, const char *path)
+{
+    switch (error) {
+        case EACCES: exec->rtl_raise(exec, "FileException.PermissionDenied", path, BID_ZERO);      break;
+        case EEXIST: exec->rtl_raise(exec, "FileException.DirectoryExists", path, BID_ZERO);       break;
+        case ENOENT: exec->rtl_raise(exec, "FileException.PathNotFound", path, BID_ZERO);          break;
+        default: {
+            char err[260];
```
ghewgill: Where the heck did you come up with a limit of 260 here?

exec/cnex/lib/file_win32.c
```diff
@@ -0,0 +1,245 @@
+#include "util.h" // Include util.h so that TRUE and FALSE are declared before including windows.h
```
ghewgill: Why is this necessary? Shouldn't the include files be able to be included in any order?

exec/cnex/lib/file_posix.c
```diff
@@ -0,0 +1,383 @@
...
+            sprintf(err, "%s: %s", path, strerror(error));
```
ghewgill: And then using an unbounded `sprintf` with arbitrary input. Stack smashing!

exec/cnex/exclude.txt
```diff
@@ -1,15 +1,15 @@
...
-file-exists.neon           # file
-file-filecopied.neon       # file
-file-filecopied1.neon      # file
-file-filecopied2.neon      # file
-file-filecopied3.neon      # file
-file-symlink.neon          # file
-file-test.neon             # file
-file-writebytes.neon       # file
-file-writelines.neon       # file
+#file-exists.neon           # file
```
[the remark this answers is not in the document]
gitlarryf: Opps.  Fixed.

exec/cnex/nstring.c
```diff
@@ -195,6 +195,18 @@ TString *string_appendString(TString *s, TString *ns)
     return s;
 }
 
+TString *string_appendBuffer(TString *s, void *buff, size_t len)
```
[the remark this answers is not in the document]
gitlarryf: Perhaps; but it was meant to be able to take a pointer to anything.  I guess if I need to ever pass a const unsigned char*, I can typecast it at the call site.
Fixed.

exec/cnex/nstring.c
gitlarryf: Changed.

exec/cnex/nstring.c
gitlarryf: Added pre-check to both functions.

exec/cnex/test_string_support.c
gitlarryf: Actually, that little bit of code shouldn't even really be there.  There are tests for the rest of these calls already.  So, I'll remove it.

exec/cnex/lib/file.c
gitlarryf: Fixed to check against bytes->string->length now.

exec/cnex/lib/file_posix.c
gitlarryf: Eww.  Yea, that is a little small.  Raised to 1024.

exec/cnex/lib/file_posix.c
gitlarryf: Fixed.

exec/cnex/lib/file.c
```diff
@@ -0,0 +1,120 @@
...
+
+    Cell *r = cell_createArrayCell(0);
+    for (;;) {
+        Cell l; cell_resetCell(&l); l.type = cString;
+
+        char line[16384];
+        if (fgets(line, sizeof(line), f) == NULL) {
+            break;
+        }
+        size_t len = strlen(line);
```
ghewgill: What happens if `len` is zero? Are you certain that a non-NULL return from `fgets` always puts a nonempty string in `line`?

exec/cnex/lib/file.c
```diff
@@ -0,0 +1,120 @@
...
+    char *filename = string_asCString(peek(exec->stack, 1)->string);
+
+    FILE *f = fopen(filename, "w");
+    if (!f) {
+        exec->rtl_raise(exec, "FileException.Open", filename, BID_ZERO);
+        goto bail;
+    }
+    for (int i = 0; i < lines->array->size; i++) {
+        size_t written = fwrite(lines->array->data[i].string->data, sizeof(char), lines->array->data[i].string->length, f);
+        if (written == 0 && ferror(f)) {
```
ghewgill: What happens if a partial write is returned here? Why is it necessary to call `ferror()`?

exec/cnex/lib/file.c
```diff
@@ -0,0 +1,120 @@
...
+            exec->rtl_raise(exec, "FileException.Write", filename, BID_ZERO);
+            goto bail;
+        }
+        fwrite("\n", sizeof(char), 1, f); // Write line-ending for each elemebnt in the array.
```
ghewgill: Spelling ("element").

exec/cnex/lib/file_posix.c
```diff
@@ -0,0 +1,383 @@
...
+
+static void handle_error(TExecutor *exec, int error, const char *path)
+{
+    switch (error) {
+        case EACCES: exec->rtl_raise(exec, "FileException.PermissionDenied", path, BID_ZERO);      break;
+        case EEXIST: exec->rtl_raise(exec, "FileException.DirectoryExists", path, BID_ZERO);       break;
+        case ENOENT: exec->rtl_raise(exec, "FileException.PathNotFound", path, BID_ZERO);          break;
+        default: {
+            char err[1024];
+            snprintf(err, 1024, "%s: %s", path, strerror(error));
```
ghewgill: Isn't there a constant size you can use here instead of hard-coding a specific numeric value? Also, use `sizeof(err)` in the `snprintf` call so you don't have to repeat the size.

exec/cnex/lib/file_win32.c
```diff
@@ -0,0 +1,243 @@
...
+    switch (error) {
+        case ERROR_ALREADY_EXISTS:      exec->rtl_raise(exec, "FileException.DirectoryExists", TCSTR(path), number_from_uint32(error));     break;
+        case ERROR_ACCESS_DENIED:       exec->rtl_raise(exec, "FileException.PermissionDenied", TCSTR(path), number_from_uint32(error));    break;
+        case ERROR_PATH_NOT_FOUND:      exec->rtl_raise(exec, "FileException.PathNotFound", TCSTR(path), number_from_uint32(error));        break;
+        case ERROR_FILE_EXISTS:         exec->rtl_raise(exec, "FileException.Exists", TCSTR(path), number_from_uint32(error));              break;
+        case ERROR_PRIVILEGE_NOT_HELD:  exec->rtl_raise(exec, "FileException.PermissionDenied", TCSTR(path), number_from_uint32(error));    break;
+        default:
+        {
+            char err[1024];
+            snprintf(err, 1024, "%s: %d", TCSTR(path), error);
```
ghewgill: Same comment about using a constant and not repeating the size value.

exec/cnex/lib/file_win32.c
```diff
@@ -0,0 +1,243 @@
...
+    t.type = cNumber; t.number = number_from_uint64(((uint64_t)(fd.nFileSizeHigh) << 32) | fd.nFileSizeLow);
+    cell_arrayAppendElement(r, t);
+    // readable: Boolean
+    t.type = cBoolean; t.boolean = TRUE;
+    cell_arrayAppendElement(r, t);
+    // writable: Boolean
+    t.type = cBoolean; t.boolean = ((fd.dwFileAttributes & FILE_ATTRIBUTE_READONLY) == 0);
+    cell_arrayAppendElement(r, t);
+    // executable: Boolean
+    t.type = cBoolean; t.boolean = name->length >= 4 && string_endsWithNoCase(name, &cexe);
```
ghewgill: I'm probably going to change this to always return FALSE for the executable bit on Windows. The definition of "executable" on win32 is too squishy to be useful, and furthermore it's not a property of the file *itself* but a property of the environment in which it lives.

exec/cnex/nstring.c
```diff
@@ -221,6 +233,37 @@ BOOL string_startsWith(TString *self, TString *s)
...
 
+BOOL string_endsWithNoCase(TString *self, TString *s)
+{
+    size_t i, si;
+    if (self->length < s->length) {
+        return FALSE;  // Can't possibly END with s, if self isn't even large enough to contain s.
+    }
+
+    for (si = 0, i = self->length - s->length; i < self->length; si++, i++) {
+        if (tolower(self->data[i]) != tolower(s->data[si])) {
```
ghewgill: Let me tell you a little story about this. In the Turkish language, `ı` and `i` are different letters. Similarly, `I` and `İ` are their upper case equivalents. See https://en.wikipedia.org/wiki/Dotted_and_dotless_I for more information. So if your OS locale is set to English, then "file" and "FILE" compare the same using this comparison algorithm. But if your OS locale is set to Turkish, then "file" and "FILE" are different. (In that case, "file" and "FİLE" would compare the same). There isn't anything you need to do at this point, it's just being aware.

exec/cnex/nstring.h
```diff
@@ -42,13 +42,18 @@ TString *string_appendString(TString *s, TString *ns);
 TString *string_appendCString(TString *s, const char *ns);
 TString *string_appendData(TString *s, char *buf, size_t len);
 TString *string_appendChar(TString *s, char c);
+TString *string_appendBuffer(TString *s, const char *buff, size_t len);
```
ghewgill: Inconsistent spelling (`buff` vs `buf` two lines above).

Actually, this function seems to be completely redundant as it is identical to `string_appendData`.

exec/cnex/lib/file.c
```diff
@@ -0,0 +1,120 @@
...
+    Cell *r = cell_createStringCell(0);
+    for (;;) {
+        char buf[16384];
+        size_t n = fread(buf, sizeof(char), sizeof(buf), f);
+        if (n == 0) {
+            break;
+        }
+        r->string = string_appendBuffer(r->string, buf, n);
+    }
+    // Ensure that the returned Cell is a cBytes type.  NOTE: This may change in the future.
```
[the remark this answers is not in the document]
gitlarryf: That was in reference to the fact that we're using a TString to hold the data.  I had planned to make TString UTF8 strings, which would force the use of a different storage method for anything that is a cBytes.  The comment doesn't HAVE to read like it does.  I'll change it.

exec/cnex/lib/file.c
```diff
@@ -0,0 +1,120 @@
...
+        exec->rtl_raise(exec, "FileException.Open", filename, BID_ZERO);
+        free(filename);
+        return;
+    }
+
+    Cell *r = cell_createArrayCell(0);
+    for (;;) {
+        Cell l; cell_resetCell(&l); l.type = cString;
+
+        char line[16384];
```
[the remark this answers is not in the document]
gitlarryf: Placed a TODO comment for this potential pitfall.

exec/cnex/lib/file.c
gitlarryf: Changed to check that the proper # of bytes gets written.

exec/cnex/lib/file.c
gitlarryf: Fixed.

exec/cnex/lib/file_win32.c
gitlarryf: Yea, I don't disagree with that.  This would be an easy change here.

exec/cnex/nstring.h
gitlarryf: Removed duplicate function.

exec/cnex/lib/file_posix.c
gitlarryf: Used PATH_MAX for the size of err.

exec/cnex/lib/file_win32.c
gitlarryf: Used MAX_PATH for size of err.

exec/cnex/lib/file_posix.c
```diff
@@ -0,0 +1,384 @@
...
+
+
+static void handle_error(TExecutor *exec, int error, const char *path)
+{
+    switch (error) {
+        case EACCES: exec->rtl_raise(exec, "FileException.PermissionDenied", path, BID_ZERO);      break;
+        case EEXIST: exec->rtl_raise(exec, "FileException.DirectoryExists", path, BID_ZERO);       break;
+        case ENOENT: exec->rtl_raise(exec, "FileException.PathNotFound", path, BID_ZERO);          break;
+        default: {
+            char err[PATH_MAX];
```
ghewgill: Wouldn't you also want to add some space for the error message?

exec/cnex/lib/file_win32.c
```diff
@@ -0,0 +1,243 @@
...
+{
+    switch (error) {
+        case ERROR_ALREADY_EXISTS:      exec->rtl_raise(exec, "FileException.DirectoryExists", TCSTR(path), number_from_uint32(error));     break;
+        case ERROR_ACCESS_DENIED:       exec->rtl_raise(exec, "FileException.PermissionDenied", TCSTR(path), number_from_uint32(error));    break;
+        case ERROR_PATH_NOT_FOUND:      exec->rtl_raise(exec, "FileException.PathNotFound", TCSTR(path), number_from_uint32(error));        break;
+        case ERROR_FILE_EXISTS:         exec->rtl_raise(exec, "FileException.Exists", TCSTR(path), number_from_uint32(error));              break;
+        case ERROR_PRIVILEGE_NOT_HELD:  exec->rtl_raise(exec, "FileException.PermissionDenied", TCSTR(path), number_from_uint32(error));    break;
+        default:
+        {
+            char err[MAX_PATH];
```
ghewgill: Same comment about leaving room for the extra information.

PR #261, opened 2019-11-13
Fix memory leak in cnex extension interface

gitlarryf:


exec/cnex/cnex.c
```diff
@@ -1339,7 +1339,9 @@ void exec_CALLX(TExecutor *self)
     uint32_t out_param_count = exec_getOperand(self);
 
     char modname[256];
-    snprintf(modname, 256, "%s/%sneon_%s", path_getPathOnly(self->module->source_path), LIBRARY_NAME_PREFIX, self->module->bytecode->strings[mod]->data);
+    char *mod_path = path_getPathOnly(self->module->source_path);
```
ghewgill: As we discussed, it's probably better not to split the memory management responsibility between caller and callee.

PR #262, opened 2019-11-15
Fix extension method interface in cnex

gitlarryf:


exec/cnex/module.h
```diff
@@ -7,6 +7,8 @@ struct tagTExecutor;
 typedef struct tagTModule {
     char *name;
     char *source_path;
+    char *path_only;
+    char *extenstion_path;
```
ghewgill: Spelling `extenstion`.

gitlarryf: D'oh.  Fixed.

PR #264, opened 2019-11-22
Fix uninitialized module globals bug in cnex

gitlarryf:


exec/cnex/cnex.c
```diff
@@ -193,9 +193,23 @@ int exec_run(TExecutor *self, BOOL enable_assert)
     invoke(self, self->modules[0], 0);
 
     for (unsigned int m = 0; m < self->init_count; m++) {
+        // Allocate all of the modules globals
+        TModule *mod = self->modules[self->init_order[(self->init_count - 1) - m]];
+        if (self->disassemble) {
```
ghewgill: This looks like it should be more like a "trace" option than "disassemble".

gitlarryf: Perhaps, I just don't have a `trace` option yet.  I'm considering putting one in, that would replace the disassemble option.  I'd like to save all the renaming until I do that, if possible.

gitlarryf: Removed, for now.  Will re-add in later pull request with fixed `-t` trace option.

PR #268, opened 2019-12-10
Add os module to cnex

gitlarryf:


exec/cnex/array.c
```diff
@@ -36,6 +36,29 @@ Array *array_createArrayFromSize(size_t iElements)
...
+    if (self == NULL) {
+        self = malloc(sizeof(Array));
+        if (self == NULL) {
+            fatal_error("Unable to allocate array for array expansion.");
+        }
+        self->size = 0;
+        self->data = NULL;
+    }
+
+    self->data = realloc(self->data, sizeof(Cell) * self->size + iElements);
```
ghewgill: The size parameter passed to realloc doesn't look right. Check your order of operations.

Also, Is this function intended to *add* iElements new elements to an existing array, or is it intended to *set* the number of elements to iElements (including shrinking)? I can't really be sure from this code and how it's called.

exec/cnex/array.c
```diff
@@ -47,6 +70,30 @@ void array_freeArray(Array *self)
...
+    if (self != NULL && self->data != NULL) {
+        for (size_t i = 0; i < self->size; i++) {
+            cell_clearCell(&self->data[i]);
+        }
+        free(self->data);
+    }
+    self->data = NULL;
+    self->size = 0;
+
+    array_expandArray(self, 0);
```
ghewgill: What is this call intended to do? You've already cleared the array up to line 82, there really isn't anything else to do.

exec/cnex/array.c
```diff
@@ -47,6 +70,30 @@ void array_freeArray(Array *self)
...
+}
+
+void array_removeItem(Array *self, size_t index)
+{
+    if (self != NULL && self->data != NULL) {
+        for(size_t i = index; i < self->size - 1;i++) {
```
ghewgill: If this is ever called with self->size == 0, there's going to be a major problem here because size is unsigned.

exec/cnex/lib/os_posix.c
```diff
@@ -0,0 +1,197 @@
...
+void object_releaseProcessObject(Object *o)
+{
+    if (o != NULL) {
+        assert(o->refcount > 0);
+        o->refcount--;
+
+        if (o->refcount <= 0) {
+            if (o->ptr != NULL) {
+                if (((TProcessHandle*)o->ptr)->proc != 0) {
+                    close(((TProcessHandle*)o->ptr)->proc);
```
ghewgill: This is not how you clean up pids. A pid is not a file handle, and you'll end up closing some random file handle here.

exec/cnex/lib/os_posix.c
```diff
@@ -0,0 +1,197 @@
...
+            free(o);
+        }
+    }
+}
+
+Cell *object_processObjectToString(Object *self)
+{
+    Cell *r = cell_fromCString("<PROCESS ");
+    char s[16];
+    sprintf(s, "%d", (pid_t)((TProcessHandle*)self->ptr)->proc);
```
ghewgill: Why is there a cast there, when ->proc is already a pid_t?

exec/cnex/lib/os_posix.c
```diff
@@ -0,0 +1,197 @@
...
+    if (init_closer == FALSE) {
+        atexit(closer);
+        init_closer = TRUE;
+    }
+
+    TProcessHandle *pid = malloc(sizeof(TProcessHandle));
+    pid->proc = fork();
+    if (pid->proc == 0) {
+        setpgid(0, 0);
+        array_clearArray(g_children);
```
ghewgill: There's no need to clear this array here. You're about to call exec, so any memory you might have allocated anywhere is completely irrelevant.

exec/cnex/lib/os_posix.c
ghewgill: sprintf is unsafe. Use snprintf.

exec/cnex/lib/os_win32.c
```diff
@@ -0,0 +1,155 @@
...
+            free(o);
+        }
+    }
+}
+
+Cell *object_processObjectToString(Object *self)
+{
+    Cell *r = cell_fromCString("<PROCESS ");
+    char s[16];
+    sprintf(s, "%llu", (int64_t)self->ptr);
```
ghewgill: Use snprintf (and probably a bigger buffer, is 15 digits long enough for a 64-bit number?)

exec/cnex/number.c
```diff
@@ -448,6 +448,16 @@ uint64_t number_to_uint64(Number x)
  * Number FROM functions
  */
 
+Number number_from_sint8(int8_t x)
+{
+    return bid128_from_int32(x & 0xFF);
```
ghewgill: This & 0xFF masking doesn't make sense here. Consider how negative numbers would be handled in this function.

exec/cnex/number.c
```diff
@@ -448,6 +448,16 @@ uint64_t number_to_uint64(Number x)
...
+}
+
+Number number_from_uint8(uint8_t x)
+{
+    return bid128_from_uint32(x & 0xFF);
```
ghewgill: This & 0xFF is redundant and unnecessary here.

gitlarryf: Removed.

exec/cnex/number.c
gitlarryf: Removed.

exec/cnex/lib/os_win32.c
gitlarryf: Increased to 32, and converted to snprintf().

exec/cnex/lib/os_posix.c
gitlarryf: Fixed.  Also improved the format string to use `"%d>"`, to avoid the double append string operation on lines 63 and 64.

exec/cnex/lib/os_posix.c
gitlarryf: What if you're calling exec twice in the same application?  Will it still be ok?  In fact.  Would it be better to eliminate the os_moduleInit(), routine, and just initialize the g_children array right at this point here?

exec/cnex/lib/os_posix.c
gitlarryf: Removed redundant cast.

exec/cnex/lib/os_posix.c
gitlarryf: What should I do here?  When a process is being cleaned up, should I call kill(), or is no operation whatsoever needed here?

exec/cnex/array.c
gitlarryf: This was to ensure that self->data was allocated, and not NULL.  But I suppose with all of the other code I have in the array handlers, it will always MAKE space, if there isn't any, so yea, I could see this call as being redundant.

exec/cnex/array.c
gitlarryf: This is meant to expand the array BY iElements.  Not to make the array size BE iElements when it's all over.  Corrected the math on this call.

exec/cnex/array.c
gitlarryf: Added sanity checked for size to be > 0.

exec/cnex/lib/os_posix.c
```diff
@@ -0,0 +1,192 @@
...
+#include "number.h"
+#include "object.h"
+#include "stack.h"
+
+
+
+static Array *g_children;
+
+typedef struct tagTProcessHandle {
+    pid_t proc;
```
ghewgill: Were you going to change this from `proc` back to `pid`?

exec/cnex/lib/os_posix.c
```diff
@@ -0,0 +1,192 @@
...
+            free(o);
+        }
+    }
+}
+
+Cell *object_processObjectToString(Object *self)
+{
+    Cell *r = cell_fromCString("<PROCESS ");
+    char s[16];
+    snprintf(s, 16, "%d>", ((TProcessHandle*)self->ptr)->proc);
```
ghewgill: Rather than doing this piecemeal, you could just snprintf the whole thing (`"<PROCESS %d>"`) into a buffer and then return a cell from that without having to append anything.

exec/cnex/lib/os_posix.c
```diff
@@ -0,0 +1,192 @@
...
+
+    push(exec->stack, cell_fromString(string_createCString(buf)));
+}
+
+void os_kill(TExecutor *exec)
+{
+    Object *process = top(exec->stack)->object; pop(exec->stack);
+    TProcessHandle *p = check_process(exec, process);
+
+    kill(-(pid_t)p->proc, SIGTERM);
```
ghewgill: Why is there a `(pid_t)` cast here? It's already the right type.

exec/cnex/lib/os_win32.c
```diff
@@ -0,0 +1,154 @@
...
+            free(o);
+        }
+    }
+}
+
+Cell *object_processObjectToString(Object *self)
+{
+    Cell *r = cell_fromCString("<PROCESS ");
+    char s[32];
+    snprintf(s, 32, "%llu>", (int64_t)self->ptr);
```
ghewgill: Same comment about string formatting as above.

exec/cnex/lib/os_posix.c
gitlarryf: Sorry, that got missed.  Fixed now.

exec/cnex/lib/os_posix.c
gitlarryf: Done.  And both of them still fit into a 32 byte buffer too.

exec/cnex/lib/os_posix.c
gitlarryf: Removed cast.

exec/cnex/lib/os_win32.c
gitlarryf: Done, and still fits within the 32 chars, even a full 64 bit number.

PR #269, opened 2019-12-11
Add instruction disassembler to cnex

gitlarryf:


exec/cnex/disassembly.c
```diff
@@ -0,0 +1,855 @@
...
+#endif
+
+
+static TInstructionDisassembler *g_disasm;
+
+TInstructionDisassembler *disasm_createDisassembler(void)
+{
+    TInstructionDisassembler *r = malloc(sizeof(TInstructionDisassembler));
+    if (r == NULL) {
+        fatal_error("Could not allocate memory for Instruction Dissembler.");
```
ghewgill: Typo, "dissembler" is not a word.

exec/cnex/disassembly.c
```diff
@@ -0,0 +1,855 @@
...
+    }
+    memset(r->out, 0x00, sizeof(r->out));
+    return r;
+}
+
+char *disasm_disassembleInstruction(TExecutor *exec)
```
ghewgill: This should return `const char *`.

exec/cnex/disassembly.c
```diff
@@ -0,0 +1,855 @@
...
+{
+    if (g_disasm == NULL) {
```
ghewgill: I don't understand why this needs to be a global. Why not make it a static inside this function? Then you wouldn't need the whole dynamic allocation in `disasm_createDisassembler` either. This seems a lot like making an unnecessary object out of something that only needs to be a function.

exec/cnex/disassembly.c
```diff
@@ -0,0 +1,855 @@
...
+    BOOL val = disasm->obj->code[disasm->index+1] != 0;
+    disasm->index += 2;
+    sprintf(&disasm->out[0], "PUSHB %d", val);
+}
+
+void disasm_PUSHN(TInstructionDisassembler *disasm)
+{
+    disasm->index++;
+    uint32_t val = get_vint(disasm->obj->code, disasm->obj->codelen, &disasm->index);
+    sprintf(&disasm->out[0], "PUSHN %s", TCSTR(disasm->obj->strings[val]));
```
ghewgill: Is this right? I don't really understand what the `TCSTR` macro does. Why is this different from what should be the same sort of thing on line 60?

exec/cnex/disassembly.c
```diff
@@ -0,0 +1,855 @@
...
+        case RESETC:  disasm_RESETC(disasm); break;
+        case PUSHPEG: disasm_PUSHPEG(disasm); break;
+        case JUMPTBL: disasm_JUMPTBL(disasm); break;
+        case CALLX:   disasm_CALLX(disasm); break;
+        case SWAP:    disasm_SWAP(disasm); break;
+        case DROPN:   disasm_DROPN(disasm); break;
+        case PUSHFP:  disasm_PUSHFP(disasm); break;
+        case CALLV:   disasm_CALLV(disasm); break;
+        case PUSHCI:  disasm_PUSHCI(disasm); break;
+    }
```
ghewgill: Should have a `default` case for invalid opcode. At least detect it and do *something*, because if you don't you'll just go into an infinite loop.

exec/cnex/disassembly.c
```diff
@@ -0,0 +1,855 @@
...
+    g_disasm->out[0] = '\0';
+    disasm_disassemble(g_disasm);
+    return g_disasm->out;
+}
+
+void disasm_PUSHB(TInstructionDisassembler *disasm)
+{
+    BOOL val = disasm->obj->code[disasm->index+1] != 0;
+    disasm->index += 2;
+    sprintf(&disasm->out[0], "PUSHB %d", val);
```
ghewgill: It would be better to write just `disasm->out` instead of the convoluted `&disasm->out[0]` which means the same thing.

gitlarryf: Fixed all occurrences.

exec/cnex/disassembly.c
gitlarryf: Added default case, which prints an invalid opcode message with the opcode number.

exec/cnex/disassembly.c
gitlarryf: It is.  Line 60 is wrong.  Fixed line 60.

exec/cnex/disassembly.c
gitlarryf: Created this as a static, and removed all of the dynamic allocation.  Originally, I was thinking of allowing trace output to go to other devices, like files, or sockets.  But, with shell redirection that is still possible; plus I thought I might had needed to do more with the Disassembly object, but it turned out to be easier than I imagined.

exec/cnex/disassembly.c
gitlarryf: Fixed.

exec/cnex/disassembly.c
gitlarryf: Removed function entirely.

exec/cnex/disassembly.h
```diff
@@ -0,0 +1,16 @@
...
+
+struct tagTExecutor;
+
+typedef struct tagTInstructionDisassembler {
+    struct tagTBytecode *obj;
+    unsigned int index;
+    char out[256];
+} TInstructionDisassembler;
+
+char *disasm_disassembleInstruction();
```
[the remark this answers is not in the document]
gitlarryf: Fixed.

exec/cnex/disassembly.c
```diff
@@ -0,0 +1,855 @@
...
+        case JUMPTBL: disasm_JUMPTBL(disasm); break;
+        case CALLX:   disasm_CALLX(disasm); break;
+        case SWAP:    disasm_SWAP(disasm); break;
+        case DROPN:   disasm_DROPN(disasm); break;
+        case PUSHFP:  disasm_PUSHFP(disasm); break;
+        case CALLV:   disasm_CALLV(disasm); break;
+        case PUSHCI:  disasm_PUSHCI(disasm); break;
+        default:
+            disasm->index++;
+            sprintf(disasm->out, "INVALID OPCODE %d", disasm->obj->code[disasm->index]);
```
ghewgill: This is going to print out the wrong opcode, because you've already incremented index.

exec/cnex/disassembly.c
```diff
@@ -0,0 +1,854 @@
...
+        case PUSHPEG: disasm_PUSHPEG(disasm); break;
+        case JUMPTBL: disasm_JUMPTBL(disasm); break;
+        case CALLX:   disasm_CALLX(disasm); break;
+        case SWAP:    disasm_SWAP(disasm); break;
+        case DROPN:   disasm_DROPN(disasm); break;
+        case PUSHFP:  disasm_PUSHFP(disasm); break;
+        case CALLV:   disasm_CALLV(disasm); break;
+        case PUSHCI:  disasm_PUSHCI(disasm); break;
+        default:
+            sprintf(disasm->out, "Unknown opcode: %d", disasm->obj->code[disasm->index]);
```
ghewgill: And now you're missing the index increment, which means you're back in infinite loop land. You have to increment the index *after* printing the opcode at that index.

PR #270, opened 2019-12-13
Add net module to cnex

gitlarryf:


exec/cnex/lib/socketx.h
```diff
@@ -0,0 +1,23 @@
...
+#else
+
+#include <arpa/inet.h>
+#include <netdb.h>
+#include <netinet/in.h>
+#include <sys/select.h>
+#include <sys/socket.h>
+#include <unistd.h>
+
+#define closesocket(x)      close(x);
```
ghewgill: Do not put a semicolon after a macro definition like this.

exec/cnex/array.c
```diff
@@ -70,6 +70,18 @@ void array_freeArray(Array *self)
...
 
+void array_clearArray(Array *self)
+{
+    if (self != NULL && self->data != NULL) {
+        for (size_t i = 0; i < self->size; i++) {
+            cell_clearCell(&self->data[i]);
+        }
+        free(self->data);
+    }
+    self->data = NULL;
```
ghewgill: Should this be inside the check for `self != NULL`? What if `self` actually is `NULL`?

exec/cnex/lib/net.c
```diff
@@ -0,0 +1,310 @@
...
+        addr.s_addr = INADDR_ANY;
+    } else {
+        addr.s_addr = inet_addr(address);
+        if (addr.s_addr == INADDR_NONE) {
+            struct hostent *he = gethostbyname(address);
+            if (he == NULL) {
+                // ToDo: Raise exception here, that host does not exist?
+                goto bail;
+            }
+            addr = *(struct in_addr*)he->h_addr_list;
```
ghewgill: This probably won't work, you're not using `h_addr_list` correctly.

exec/cnex/lib/socketx.h
gitlarryf: Fixed.

exec/cnex/array.c
gitlarryf: Good call!  Fixed.

exec/cnex/lib/net.c

PR #275, opened 2020-01-23
Add mmap module to cnex

gitlarryf:


exec/cnex/lib/mmap_posix.c
```diff
@@ -0,0 +1,184 @@
...
+
+void object_releaseMMapObject(Object *o)
+{
+    if (o != NULL) {
+        assert(o->refcount > 0);
+        o->refcount--;
+
+        if (o->refcount <= 0) {
+            if (o->ptr != NULL) {
+                free(o->ptr);
```
ghewgill: I can't help but think that there's something missing here. How does the file get closed if all the references to the object go away?

exec/cnex/lib/mmap_posix.c
```diff
@@ -0,0 +1,184 @@
...
+
+#include "cell.h"
+#include "exec.h"
+#include "nstring.h"
+#include "number.h"
+#include "object.h"
+#include "stack.h"
+
+
+typedef struct tagTMmapObject {
```
ghewgill: You should reconsider your use of `typedef` everywhere. It's just not necessary. All it does is (a) introduce all these odd `tagWhatever` names, and (b) save typing `struct` in some places. It's better to type `struct`.

exec/cnex/lib/mmap_posix.c
```diff
@@ -0,0 +1,184 @@
...
+    Number offset = top(exec->stack)->number; pop(exec->stack);
+    Object *pf = top(exec->stack)->object;  pop(exec->stack);
+    Cell *ret = cell_createStringCell(0);
+    ret->type = cBytes;
+
+    TMmapObject *f = check_file(exec, pf);
+
+    uint64_t o = number_to_uint64(offset);
+    if (o >= f->len) {
+        ret->type = cBytes;
```
ghewgill: Seems like a redundant assignment, since you've already done that a few lines above.

exec/cnex/lib/mmap_posix.c
```diff
@@ -0,0 +1,184 @@
...
+    Number offset = top(exec->stack)->number; pop(exec->stack);
+    Object *pf = top(exec->stack)->object;  pop(exec->stack);
+
+    TMmapObject *f = check_file(exec, pf);
+    if (f == NULL) {
+        string_freeString(data);
+        return;
+    }
+    uint64_t o = number_to_uint64(offset);
+    if (o >= f->len) {
```
ghewgill: I don't think you need both this and the following test. The first condition is covered by the second test, and the resulting action is the same. I think you only need the second one.

exec/cnex/lib/mmap_win32.c
```diff
@@ -0,0 +1,208 @@
...
+void mmap_open(TExecutor *exec)
+{
+    pop(exec->stack);
+    char *name = string_asCString(top(exec->stack)->string); pop(exec->stack);
+
+    char err[64];
+    TMmapObject *f = malloc(sizeof(TMmapObject));
+
+    f->file = INVALID_HANDLE_VALUE;
+    if (strcmp(name, "") != 0) {
```
ghewgill: Don't use `strcmp` just to check for an empty string.

exec/cnex/lib/mmap_posix.c
gitlarryf: Removed typedef, and changed all references to MmapObject, instead of TMmapObject.

exec/cnex/lib/mmap_posix.c
gitlarryf: Fixed.  If `(o->ptr)->fd > 0` it will unmap the view, and close the file.

exec/cnex/lib/mmap_posix.c
gitlarryf: Removed redundant assignment.

exec/cnex/lib/mmap_posix.c
gitlarryf: Removed `if (o >= f->len) {` test.

exec/cnex/lib/mmap_posix.c
```diff
@@ -0,0 +1,183 @@
...
+void object_releaseMMapObject(Object *o)
+{
+    if (o != NULL) {
+        assert(o->refcount > 0);
+        o->refcount--;
+
+        if (o->refcount <= 0) {
+            if (o->ptr != NULL) {
+                struct MmapObject *f = o->ptr;
+                if (f->fd > 0) {
```
ghewgill: In posix, 0 is a valid file descriptor, so you should check for `>= 0` here.

exec/cnex/lib/mmap_posix.c
```diff
@@ -0,0 +1,183 @@
...
+                free(o->ptr);
+            }
+        }
+    }
+}
+
+Cell *object_mmapObjectToString(Object *self)
+{
+    char s[32];
+    snprintf(s, 32, "<mmap %" PRId64 ">", (int64_t)((struct MmapObject*)self->ptr)->fd);
```
ghewgill: Use `sizeof(s)` here, don't repeat the number.

gitlarryf: Fixed all instances of static lengths with `sizeof()`.

PR #277, opened 2020-01-29
Cnex libs net fixup

gitlarryf:


exec/cnex/lib/net.c
```diff
@@ -1,5 +1,6 @@
 #include "net.h"
-
+// Needed for gethostbyname() on some Linux platforms.
```
ghewgill: Still need a blank line here.

gitlarryf: Done.

exec/cnex/global.c
```diff
@@ -166,7 +166,7 @@ TDispatch gfuncDispatch[] = {
     PDFUNC("net$socket_send",           net_socket_send),
     PDFUNC("net$socket_select",         net_socket_select),
 
-    // OS - Operating System function calls
+ // OS - Operating System function calls
```
ghewgill: The only thing left in this commit is screwing up the indenting, which is fixed in the next commit anyway. Please remove this unneeded commit.

PR #279, opened 2020-02-14
Enhance use of cJSON in cnex

gitlarryf:


exec/cnex/cnex.c
```diff
@@ -381,7 +383,7 @@ void exec_raiseLiteral(TExecutor *self, TString *name, TString *info, Number cod
                 cJSON_ArrayForEach(symbols, lines)
                 {
                     jip = cJSON_GetArrayItem(symbols, 0);
-                    if ((uint64_t)jip->valueint == p) {
+                     if (!cJSON_IsInvalid(jip) && (uint64_t)jip->valueint == p) {
```
ghewgill: Somehow this got misaligned with an extra leading space.

gitlarryf: Good catch, there!  Fixed.

PR #286, opened 2020-03-18
Add SQLite module to cnex

gitlarryf:


exec/cnex/lib/sqlite.c
```diff
@@ -0,0 +1,382 @@
...
+static Dictionary *Cursors;
+
+void sqlite_initModule(void)
+{
+    Cursors = dictionary_createDictionary();
+}
+
+void sqlite_shutdownModule(void)
+{
+    dictionary_freeDictionary(Cursors);
```
ghewgill: Do you want to actually close any cursors that might be left in this dictionary, or not worry about it?

exec/cnex/lib/sqlite.c
```diff
@@ -0,0 +1,382 @@
...
+
+
+void object_releaseDatabaseObject(Object *o)
+{
+    if (o != NULL) {
+        assert(o->refcount > 0);
+        o->refcount--;
+        if (o->refcount <= 0) {
+            if (o->ptr != NULL) {
+                assert(sqlite3_close(o->ptr) == SQLITE_OK);
```
ghewgill: Never ever put a statement with side effects like this inside `assert()`. What happens when you build without asserts enabled? The close function will never be called. Save the result in a variable and assert on the value of the result.

exec/cnex/lib/sqlite.c
```diff
@@ -0,0 +1,382 @@
...
+    r->stmt = NULL;
+    r->columns = 0;
+    r->exec = exec;
+
+    return r;
+}
+
+static int callback(void *rowscell, int columns, char **values, char **names)
+{
+    (void)names; // prevent compiler warning "expected formal parameter list, not a type list"
```
ghewgill: This comment doesn't make sense. Isn't this statement here to indicate an unused parameter?

exec/cnex/lib/sqlite.c
```diff
@@ -0,0 +1,382 @@
...
+    free(name);
+}
+
+void sqlite_exec(TExecutor *exec)
+{
+    Cell *parameters = cell_fromCell(top(exec->stack)); pop(exec->stack);
+    char *sql = string_asCString(top(exec->stack)->string); pop(exec->stack);
+    Cell *pdb = cell_fromCell(top(exec->stack)); pop(exec->stack);
+
+    Object *db = check_database(exec, pdb->object);
```
ghewgill: Shouldn't you check the return value of this function here?

exec/cnex/object.h
```diff
@@ -45,6 +46,7 @@ Object *object_createStringObject(TString *s);
 Object *object_createFileObject(FILE *f);
 Object *object_createProcessObject(void *h);
 Object *object_createMMapObject(struct MmapObject *m);
+Object *object_createDatabaseObject(struct sqlite3 *db);
```
ghewgill: There's no need to have this declaration here, there does not need to be any changes to `object.h` at all.

exec/cnex/lib/sqlite.c
gitlarryf: I wasn't sure if I should worry about it, or make it the programmers responsibility.  I think the code will exception on close if you have a cursor that is left open.  I'll test it, try it and see how it runs.

exec/cnex/lib/sqlite.c
gitlarryf: Removed.  Also removed the assertion in general.  If the programmer fails to close the DB properly, then we cannot be responsible for trying to clean it up for them.

exec/cnex/lib/sqlite.c
gitlarryf: Corrected.

exec/cnex/lib/sqlite.c
gitlarryf: Changed all references to check if db is null, and return from the function right away.

exec/cnex/object.h
gitlarryf: Removed.

exec/cnex/lib/sqlite.c
```diff
@@ -0,0 +1,417 @@
...
+    if (db == NULL) {
+        cell_freeCell(parameters);
+        free(sql);
+        cell_freCell(pdb);
```
ghewgill: This can't possibly have worked?

exec/cnex/lib/sqlite.c
```diff
@@ -0,0 +1,417 @@
...
+    free(sql);
+    cell_freeCell(parameters);
+    cell_freeCell(pdb);
+}
+
+void sqlite_execOne(TExecutor *exec)
+{
+    Cell *parameters = cell_fromCell(top(exec->stack)); pop(exec->stack);
+    char *sql = string_asCString(top(exec->stack)->string); pop(exec->stack);
+    Cell *pdb = cell_fromCell(top(exec->stack)); pop(exec->stack);
```
ghewgill: Do you really need to make a *copy* of the cell here? I can't even see where you free this in the normal case.

exec/cnex/object.h
```diff
@@ -9,6 +9,7 @@
 #include "util.h"
 
 struct MmapObject;
+struct sqlite3;
```
ghewgill: Why is this still here? You don't need ANY changes to object.h.

exec/cnex/lib/sqlite.c
gitlarryf: Woah.  Now THAT was a good one!  I must have hit it after compiling/testing, but before adding everything to git.  Fixed.

exec/cnex/lib/sqlite.c
gitlarryf: It's not MANDATORY that I make a copy, but if I don't, then the code starts to look weird.  Like.
`Cell *pdb = top(exec->stack);
Object *db = check_database(exec, pdb->object);
pop(exec->stack);
if (db == NULL)
...
`
So, I just opted to copy the cell.  I'm not thrilled with either solution, so I'll take your direction on which method you think is best.  Based on this suggestion, I'd like to go back over all the instances that use this style of construction and redo them, so the code is the same in all places.

gitlarryf: Actually.  Now that I look at what I wrote above, I think I'm going to attack it from a different perspective all together.  I'm going to eliminate the Cell altogether.  Don't really need it.

exec/cnex/object.h
gitlarryf: Removed

exec/cnex/lib/sqlite.c
gitlarryf: Fixed.  Used a new approach.

exec/cnex/lib/sqlite.c
```diff
@@ -0,0 +1,404 @@
+#ifdef _WIN32
+#define _CRT_SECURE_NO_WARNINGS
```
ghewgill, line 2: I'm curious about the placement of this `#define`. Is this something that needs to be defined before including `sqlite.h`? Or is it something that just relates to the code within this .cpp file? I suspect that it is the latter, in which case it should be moved *below* all the includes. That makes sure that the define only affects *this* source file, and not the included headers.

exec/cnex/lib/sqlite.c
```diff
@@ -0,0 +1,404 @@
...
+            }
+            free(o);
+        }
+    }
+}
+
+Cell *object_databaseObjectToString(Object *self)
+{
+    char s[32];
+    snprintf(s, 32, "<SQLITE %" PRId64 ">", (int64_t)self->ptr);
```
ghewgill, line 57: Use `sizeof(s)` here instead of repeating the `32` value

exec/cnex/lib/sqlite.c
```diff
@@ -0,0 +1,404 @@
...
+        for (int i = 0; i < columns; i++) {
+            cell_arrayAppendElementPointer(result, cell_fromCString((const char *)sqlite3_column_text(stmt, i)));
+        }
+    } else {
+        exec->rtl_raise(exec, "SqlException", sqlite3_errmsg(db->ptr), number_from_sint32(r));
+        goto done;
+    }
+    Retval = TRUE;
+
+done:
```
ghewgill, line 274: There's considerable inconsistency in the naming of these goto labels. The previous function uses `cleanup`, this one uses `done`, and the next function doesn't use one at all (but it could). Suggest making these consistent.

exec/cnex/lib/sqlite.c
```diff
@@ -0,0 +1,404 @@
...
+        free(sql);
+        return;
+    }
+
+    Cell *rows = cell_createArrayCell(0);
+    char *errmsg = NULL;
+    int r = sqlite3_exec(db->ptr, sql, callback, rows, &errmsg);
+    if (r != SQLITE_OK) {
+        exec->rtl_raise(exec, "SqlException", sqlite3_errmsg(db->ptr), number_from_sint32(r));
+    }
```
ghewgill, line 298: I think you need to bail out after calling `rtl_raise` do you don't push anything else onto the stack after that.

exec/cnex/lib/sqlite.c
```diff
@@ -0,0 +1,404 @@
...
+
+void sqlite_cursorClose(TExecutor *exec)
+{
+    Cell *name = cell_fromCell(top(exec->stack)); pop(exec->stack);
+
+    Cell *c = dictionary_findDictionaryEntry(Cursors, name->string);
+    if (c == NULL) {
+        exec->rtl_raise(exec, "SqlException", TCSTR(name->string), BID_ZERO);
+        cell_freeCell(name);
+        return;
```
ghewgill, line 384: Would you like to use a `goto` label here to avoid duplicating the cleanup code?

exec/cnex/lib/sqlite.c
gitlarryf: Yea, this is really only for the object::toString() function.  So, I moved it to after all of the includes.

exec/cnex/lib/sqlite.c
gitlarryf: Fixed.

exec/cnex/lib/sqlite.c
gitlarryf: I tried to only use them when necessary, and name them for what they actually were in there for.  Would it be bad to use a "bail" tag to cover all cases?

exec/cnex/lib/sqlite.c
gitlarryf: Yep.  Fixed.

exec/cnex/lib/sqlite.c
gitlarryf: Added a goto here.

exec/cnex/lib/sqlite.c
```diff
@@ -0,0 +1,407 @@
...
+void sqlite_close(TExecutor *exec)
+{
+    Object *db = check_database(top(exec->stack)); pop(exec->stack);
+
+    if (db == NULL) {
+        exec->rtl_raise(exec, "SqliteException.InvalidDatabase", "", BID_ZERO);
+        return;
+    }
+
+    int retval = sqlite3_close(db->ptr);
```
ghewgill, line 316: According to https://sqlite.org/c3ref/close.html, the `sqlite3_close_v2` function would probably be more appropriate here. As a bonus you won't get the `SQLITE_BUSY` return value.

exec/cnex/lib/sqlite.c
ghewgill: Did you? I don't see one.

PR #287, opened 2020-04-29
Add posix module to cnex

gitlarryf:


exec/cnex/global.c
```diff
@@ -174,6 +175,80 @@ TDispatch gfuncDispatch[] = {
     PDFUNC("os$system",                 os_system),
     PDFUNC("os$wait",                   os_wait),
 
+    // posix - posix OS functions
+    PDFUNC("posix$access",              posix_access),
+    PDFUNC("posix$chdir",               posix_chdir),
+    PDFUNC("posix$chmod",               posix_chmod),
+    PDFUNC("posix$chown",               posix_chown),
+    PDFUNC("posix$close",               posix_Close),
```
ghewgill, line 183: Why is "Close" uppercased here? Is that a typo or does it conflict with something else called "posix_close"?

exec/cnex/lib/posix.c
```diff
@@ -0,0 +1,781 @@
...
+
+void posix_getgid(struct tagTExecutor *exec)
+{
+    push(exec->stack, cell_fromNumber(number_from_sint32(getgid())));
+}
+
+void posix_gethostname(struct tagTExecutor *exec)
+{
+    char name[256];
+    Cell *r = wrap(gethostname(name, 256));
```
ghewgill, line 252: Use `sizeof(name)` here instead of repeating "256".

exec/cnex/lib/posix.c
```diff
@@ -0,0 +1,781 @@
...
+void posix_isatty(struct tagTExecutor *exec)
+{
+    Number fildes = top(exec->stack)->number; pop(exec->stack);
+
+    push(exec->stack, cell_fromBoolean(isatty(number_to_sint32(fildes)) != 0));
+}
+
+void posix_kill(struct tagTExecutor *exec)
+{
+    Number pid = top(exec->stack)->number; pop(exec->stack);
```
ghewgill, line 315: Are you sure these parameters are popped in the correct order?

exec/cnex/lib/posix.c
```diff
@@ -0,0 +1,781 @@
...
+    int fds[2];
+    int r = pipe(fds);
+    if (r < 0) {
+        push(exec->stack, wrap(r));
+        return;
+    }
+    Cell *rfd = cell_fromNumber(number_from_sint32(fds[0]));
+    Cell *wfd = cell_fromNumber(number_from_sint32(fds[1]));
+
+    push(exec->stack, wrap(r));
```
ghewgill, line 434: This call to `wrap()` is not necessary here, because you've already checked the return value above.

exec/cnex/lib/posix.c
```diff
@@ -0,0 +1,781 @@
...
+    Cell *r = wrap(readlinkat(number_to_sint32(fd), path, buf->string->data, n));
+
+    push(exec->stack, buf);
+    push(exec->stack, r);
+}
+#endif
+
+void posix_rename(struct tagTExecutor *exec)
+{
+    char *new_ = string_asCString(top(exec->stack)->string); pop(exec->stack);
```
ghewgill, line 512: Why do these local variables have the weird trailing underscores?

exec/cnex/lib/posix.c
```diff
@@ -0,0 +1,781 @@
...
+
+    push(exec->stack, wrap(symlinkat(name1, number_to_sint32(fd), name2)));
+    free(name1);
+    free(name2);
+}
+#endif
+
+void posix_sync(struct tagTExecutor *exec)
+{
+    assert(exec);
```
ghewgill, line 647: What is the purpose of this assert?

exec/cnex/global.c
gitlarryf: Yes.  It conflicted, so the easy solution was to ust uppercase it in the code.

exec/cnex/lib/posix.c
gitlarryf: Fixed.

exec/cnex/lib/posix.c
gitlarryf: No.  After looking at posix.cpp, I do believe I had these in the wrong order.  Fixed.

exec/cnex/global.c
ghewgill: What did it conflict with?

exec/cnex/lib/posix.c
gitlarryf: Fixed.

exec/cnex/lib/posix.c
gitlarryf: I had modeled my code off posix.cpp; which might not have been the smartest thing to go, since we're in very different language universes.  Fixed.

exec/cnex/lib/posix.c
gitlarryf: That was basically because the unused exec parameter.  So, I figured an assert to assert that the value isn't NULL would have been ok here.  I can change it to simply be:
`exec=exec; // unused variable`

ghewgill: A dummy assignment like that would be better, because that would actually work. An assert won't (because when you turn off asserts, the code goes away and the warning would come back).

exec/cnex/global.c
gitlarryf: I don't entirely know.  Looking back, I think it was an attempt to avoid any naming conflicts, but I don't see it as necessary now.

exec/cnex/lib/posix.c
```diff
@@ -0,0 +1,781 @@
+#include "posix.h"
+
+#define _XOPEN_SOURCE       700
```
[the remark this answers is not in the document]
gitlarryf: Fixed.

PR #318, opened 2020-10-10
Add basic memory management to cnex

gitlarryf:
This build adds a very basic garbage collector to cnex, as well as implementing a build option that allows leak detection on all test files.

### DIFF
--- a/exec/cnex/cnex.c
+++ b/exec/cnex/cnex.c
@@ -2,6 +2,7 @@
 #define _CRTDBG_MAP_ALLOC
 #include <stdlib.h>
 #include <crtdbg.h>
+#include <windows.h>
 #endif
 // The following is needed for strdup on Posix platforms.
 #define _POSIX_C_SOURCE 200809L
@@ -26,6 +27,7 @@
 #include "dictionary.h"
 #include "disassembly.h"
 #include "exec.h"
+#include "gc.h"
 #include "global.h"
 #include "framestack.h"
 #include "module.h"
@@ -71,8 +73,12 @@ void exec_freeExecutor(TExecutor *e)
     if (e->exit_code == 0) {
         assert(isEmpty(e->stack));
     }
+    // Next, destroy the opstack.
     destroyStack(e->stack);
     free(e->stack);
+
+    // Next, free all Neon-allocated objects.
+    heap_freeHeap(e);
 
     // Next, iterate all allocated modules, freeing them as we go.
     for (unsigned int m = e->module_count; m > 0; m--) {
@@ -143,6 +149,13 @@ BOOL ParseOptions(int argc, char* argv[])
 
 int main(int argc, char* argv[])
 {
+#ifdef __MS_HEAP_DBG
+    /* Used for debugging cnex, looking for memory leaks. */
+    _CrtSetDbgFlag(_CRTDBG_ALLOC_MEM_DF | _CRTDBG_LEAK_CHECK_DF);
+    if (IsDebuggerPresent()) {
+        _CrtSetBreakAlloc(0);  // Can be set to break on specific allocation.
+    }
+#endif
     int ret = 0;
     gOptions.pszExecutableName = path_getFileNameOnly(argv[0]);
 
@@ -172,6 +185,8 @@ int main(int argc, char* argv[])
                         "CallStack Height       : %" PRId32 "\n"
                         "Global Size            : %" PRIu32 "\n"
                         "Max Framesets          : %d\n"
+                        "Max Allocations        : %" PRIu64 "\n"
+                        "Remaining Objects      : %" PRIu64 "\n"
                         "Execution Time         : %fms\n",
                         g_executor->diagnostics.total_opcodes,
                         g_executor->stack->max + 1,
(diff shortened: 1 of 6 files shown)


exec/cnex/cnex.c
```diff
@@ -172,6 +185,8 @@ int main(int argc, char* argv[])
                         "CallStack Height       : %" PRId32 "\n"
                         "Global Size            : %" PRIu32 "\n"
                         "Max Framesets          : %d\n"
+                        "Max Allocations        : %" PRIu64 "\n"
+                        "Remaining Objects      : %u\n"
```
ghewgill, line 189: Why are these two values above different types? For the allocator that doesn't free until it's done, wouldn't they always be the same number?

exec/cnex/cnex.c
```diff
@@ -1388,8 +1418,15 @@ void exec_ALLOC(TExecutor *self)
     self->ip++;
     uint32_t val = exec_getOperand(self);
 
-    Cell *cell = cell_createArrayCell(val);
+    Cell *cell = heap_allocObject(self);
+    cell->array = array_createArrayFromSize(val);
+    cell->type = cArray;
     push(self->stack, cell_fromAddress(cell));
+    self->allocations++;
```
ghewgill, line 1425: Where does `allocations` get reset to zero?

exec/cnex/gc.c
```diff
@@ -0,0 +1,73 @@
...
+    }
+    exec->diagnostics.total_allocations++;
+
+    Cell *p = cell_newCell();
+    o->allocNum = exec->diagnostics.total_allocations;
+    o->object = p;
+    o->next = NULL;
+
+    // Push newaly allocated object on the virtual heap.
+    if (exec->firstObject != NULL) {
```
ghewgill, line 32: You don't have to check to see whether `exec->firstObject` is null or not before assigning it. Also, since you don't need to check here, you don't need the assignment on line 29 either.

exec/cnex/gc.c
```diff
@@ -0,0 +1,73 @@
...
+    return p;
+}
+
+unsigned int heap_getObjectCount(TExecutor *exec)
+{
+    // Count all allocated objects, regardless if they are in use or not.
+    unsigned int ret = 0;
+    TAllocationObject *object = exec->firstObject;
+
+    while (object) {
```
ghewgill, line 46: In this case, use `object != NULL` for better readability.

exec/cnex/gc.h
```diff
@@ -0,0 +1,21 @@
+#ifndef GC_H
+#define GC_H
+#include <stddef.h>
+#include <stdint.h>
+#include <util.h>
```
ghewgill, line 5: I don't think you need any of the above three includes in this file.

ghewgill (changes requested):

exec/cnex/cnex.c
gitlarryf: In this version, yes; they would.

exec/cnex/cnex.c
gitlarryf: Never.  That's the allocation count.  When this supports an actual collector, then it will reset when it sweeps.

exec/cnex/gc.c
gitlarryf: Fixed in 3 locations.

exec/cnex/gc.c
gitlarryf: Fixed logic around this.

exec/cnex/cnex.c
```diff
@@ -192,7 +209,15 @@ int main(int argc, char* argv[])
     ext_cleanup();
 
     free(gOptions.pszExecutablePath);
-
+#ifdef __MS_HEAP_DBG
+    /* If you're doing HEAP tracking, and cnex leaks memory, then 
+       _CrtDempMemoryLeaks() will return true, to ensure that the 
+       test that is being debugged will fail.  (This will also dump
+       all of the allocations that were never free()'ed.)       */
+    if (_CrtDumpMemoryLeaks() == TRUE) {
```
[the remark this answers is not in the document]
gitlarryf: Removed.

exec/cnex/cnex.c
```diff
@@ -2,6 +2,7 @@
 #define _CRTDBG_MAP_ALLOC
 #include <stdlib.h>
 #include <crtdbg.h>
+#include <Windows.h>
```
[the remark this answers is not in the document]
gitlarryf: Actually, it **is** Windows.h, but Windows is case insensitive, so I changed it.

exec/cnex/cnex.c
gitlarryf: Changed All object counters to `size_t`.

exec/cnex/cnex.c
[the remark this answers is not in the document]
ghewgill: Huh, I never noticed that. Oh well, I think I've always seen people write `<windows.h>`, regardless of the actual case of the file name.

exec/cnex/gc.h
```diff
@@ -0,0 +1,19 @@
+#ifndef GC_H
+#define GC_H
+#include <stdint.h>
```
ghewgill, line 3: I'm not entirely sure why, but on my system the `stdint.h` header does not declare `size_t`. But `stddef.h` for example does. The documentation says a bunch of choices *should* work: https://en.cppreference.com/w/c/types/size_t

Since you need `stdint.h` for `uint64_t`, I suggest including both.

ghewgill (changes requested):

exec/cnex/gc.h
gitlarryf: Fixed.

PR #326, opened 2020-11-06
Fix number_from_string() function in cnex

gitlarryf:


ghewgill:
This also needs to include the corresponding test.

ghewgill (changes requested):
This also needs to include the corresponding test.

exec/cnex/number.c
```diff
@@ -154,6 +154,55 @@ void number_toString(Number x, char *buf, size_t len)
 
 Number number_from_string(char *s)
```
ghewgill, line 155: This should be `const char *`.

ghewgill (changes requested):

exec/cnex/number.c
gitlarryf: Changed.

PR #327, opened 2020-11-12
Fix win32 random module

gitlarryf:


exec/cnex/lib/random_win32.c
```diff
@@ -19,17 +19,23 @@ void random_bytes(TExecutor *exec)
...
+            exec->rtl_raise(exec, "RandomException", err);
+            return;
+        }
     }
     char *p = r->string->data;
     if (!CryptGenRandom(hCryptProv, (DWORD)count, (BYTE*)p)) {
         cell_freeCell(r);
         CryptReleaseContext(hCryptProv, 0);
-        exec->rtl_raise(exec, "RandomException", "failed to get random data");
+        snprintf(err, sizeof(err), "failed to get random data. Error: 0x%08x", GetLastError());
```
ghewgill, line 37: You will need to capture `GetLastError()` immediately after the call to `CryptGenRandom()` before you do anything else. Otherwise you risk overwriting the last error value and having a false report.

exec/cnex/lib/random_win32.c
```diff
@@ -19,17 +19,23 @@ void random_bytes(TExecutor *exec)
...
+    char err[64];
     HCRYPTPROV hCryptProv;
     if (!CryptAcquireContext(&hCryptProv, NULL, NULL, PROV_RSA_FULL, 0)) {
-        cell_freeCell(r);
-        exec->rtl_raise(exec, "RandomException", "could not acquire default crypto context");
-        return;
+        // No default context/container was found, so we'll attempt to just create one.
+        if (!CryptAcquireContext(&hCryptProv, NULL, NULL, PROV_RSA_FULL, CRYPT_NEWKEYSET)) {
+            cell_freeCell(r);
+            snprintf(err, sizeof(err), "could not create crypto context. Error: 0x%08x", GetLastError());
```
ghewgill, line 28: Same as below about `GetLastError()`.

lib/random_win32.cpp
```diff
@@ -13,17 +13,21 @@ std::vector<unsigned char> bytes(Number count)
...
+            snprintf(err, sizeof(err), "could not create crypto context. Error 0x%08x", GetLastError());
+            throw RtlException(Exception_RandomException, utf8string(err));
+        }
     }
     unsigned char *p = r.data();
     if (!CryptGenRandom(hCryptProv, static_cast<DWORD>(icount), p)) {
         CryptReleaseContext(hCryptProv, 0);
-        // Should we return GetLastError() in the exception too?
-        throw RtlException(Exception_RandomException, utf8string("failed to get random data"));
+        snprintf(err, sizeof(err), "failed to get random data. Error: 0x%08x", GetLastError());
```
ghewgill, line 29: Same as above about `GetLastError()`.

ghewgill (changes requested):

exec/cnex/lib/random_win32.c
gitlarryf: Bah.  Dumb mistake.  Fixed in all locations.

PR #333, opened 2020-12-17
Add shebang parsing to cnex

gitlarryf:


exec/cnex/bytecode.c
```diff
@@ -109,7 +109,11 @@ void bytecode_loadBytecode(TBytecode *b, const uint8_t *bytecode, unsigned int l
 
     assert(b != NULL);
 
-    if (memcmp(bytecode, "Ne\0n", 4) != 0) {
+    if (memcmp(bytecode, "#!", 2) == 0) {
+        while (bytecode[i] != '\n' && i < len) i++;
```
ghewgill, line 113: This needs to compare `i < len` *before* accessing the array. Also, use curly braces even though you've only got an increment statement.

ghewgill (changes requested):
This needs test support.

exec/cnex/bytecode.c
```diff
@@ -109,7 +109,16 @@ void bytecode_loadBytecode(TBytecode *b, const uint8_t *bytecode, unsigned int l
 
     assert(b != NULL);
 
-    if (memcmp(bytecode, "Ne\0n", 4) != 0) {
+    if (memcmp(bytecode, "#!", 2) == 0) {
+        while (i < len && bytecode[i] != '\n') {
+            i++;
+        }
+        if (i+1 >= len) {
```
ghewgill, line 116: Instead of using `i+1` here and incrementing after, it would be cleaner to increment first, then check `i >= len`.

ghewgill (changes requested):
This still needs a test to actually test this functionality.

exec/cnex/bytecode.c
gitlarryf: Fixed.

exec/cnex/bytecode.c
gitlarryf: Fixed.

PR #334, opened 2020-12-17
Add native object support to cnex

gitlarryf:


exec/cnex/lib/runtime_win32.c
```diff
@@ -0,0 +1,389 @@
+#include "runtime.h"
+
+#define COBJMACROS
+
+#include <iso646.h>
```
ghewgill, line 5: Why include this header file if you're not going to use anything in it?

exec/cnex/lib/runtime_win32.c
```diff
@@ -0,0 +1,389 @@
...
+#include <windows.h>
+
+#include "assert.h"
+#include "cell.h"
+#include "enums.h"
+#include "exec.h"
+#include "object.h"
+#include "stack.h"
+
+#ifdef _WIN32
```
ghewgill, line 15: I think you can already assume that this file is being built for win32, no need for an ifdef.

exec/cnex/lib/runtime_win32.c
```diff
@@ -0,0 +1,389 @@
...
+}
+
+// ToDo: Implement UTF8 strings!
+char *utf8string_from_bstr(BSTR bstr)
+{
+    if (bstr == NULL) {
+        return "null";
+    }
+    int nchars = WideCharToMultiByte(CP_ACP, 0, bstr, -1, NULL, 0, NULL, NULL);
+    char *buf = malloc(nchars);
```
ghewgill, line 48: Memory leak here (not freed).

exec/cnex/lib/runtime_win32.c
```diff
@@ -0,0 +1,389 @@
...
+    r = com->lpVtbl->Invoke(com, dispid, &IID_NULL, GetUserDefaultLCID(), DISPATCH_METHOD, &params, &rv, &ei, &argerr);
+    for (size_t i = 0; i < pargs->size; i++) {
+        VariantClear(&params.rgvarg[i]);
+    }
+    free(params.rgvarg);
+    if (!handleInvokeResult(exec, r, params.cArgs, &ei, argerr)) {
+        return FALSE;
+    }
+
+    ((Cell*)*result)->object = ObjectFromVariant(exec, &rv);
```
ghewgill, line 246: There should not need to be a cast to `(Cell*)` here, because that's what the pointer already is. Remove the casts.

ghewgill (changes requested):

exec/cnex/lib/runtime_win32.c
gitlarryf: Fixed.

exec/cnex/lib/runtime_win32.c
gitlarryf: Fixed.

exec/cnex/lib/runtime_win32.c
gitlarryf: It is free'd by the caller.  Fixed an instance where it wasn't free'ed.

exec/cnex/lib/runtime_win32.c
gitlarryf: Fixed.

PR #336, opened 2021-01-05
csnex initial commit

gitlarryf:


exec/csnex/Bytecode.cs
```diff
@@ -0,0 +1,144 @@
...
+        public string source_path;
+        private byte[] source_hash;
+        private uint version;
+        public uint global_size;
+        public List<String> strtable;
+        private uint strtablesize;
+        public byte[] code;
+        public uint codelen;
+
+        private uint functionsize;
```
ghewgill, line 18: This does not seem to need to be a class member.

exec/csnex/Bytecode.cs
```diff
@@ -0,0 +1,144 @@
...
+    {
+        private byte[] obj;
+        public string source_path;
+        private byte[] source_hash;
+        private uint version;
+        public uint global_size;
+        public List<String> strtable;
+        private uint strtablesize;
+        public byte[] code;
+        public uint codelen;
```
ghewgill, line 16: This doesn't seem to be required. The length of the code is just the length of the `code` array, you don't have to keep track of the length separately.

exec/csnex/Bytecode.cs
```diff
@@ -0,0 +1,144 @@
...
+{
+    public class Bytecode
+    {
+        private byte[] obj;
+        public string source_path;
+        private byte[] source_hash;
+        private uint version;
+        public uint global_size;
+        public List<String> strtable;
+        private uint strtablesize;
```
ghewgill, line 14: This does not seem to need to be a class member.

exec/csnex/CMakeLists.txt
```diff
@@ -0,0 +1,21 @@
+set(CSNEX
```
ghewgill, line 1: This variable would be better named `CSNEX_SOURCES` (yes I know this was my mistake).

exec/csnex/Extensions.cs
```diff
@@ -0,0 +1,18 @@
+﻿/*****************************************************************************
+ *   File: Extensions.cs
+ * Author: Larry Frieson
+ *   Date: 2020-12-24 15:30:08
+ *   Desc: This file contains only a single class, used to extend the
+ *         functionalry of .NET core library objects.
```
ghewgill, line 6: Spelling

exec/csnex/Extensions.cs
```diff
@@ -0,0 +1,18 @@
...
+ ****************************************************************************/
+internal static partial class CodeExtensions
+{
+    public static byte[] CopyFrom(this byte[] source, uint start, int len)
```
ghewgill, line 10: Why is this function necessary when the Array.Copy method (https://docs.microsoft.com/en-us/dotnet/api/system.array.copy?view=net-5.0) can already do this?

exec/csnex/Opcode.cs
```diff
@@ -0,0 +1,111 @@
+﻿namespace csnex
+{
+    public class NeonOpcode
```
ghewgill, line 3: What is the purpose of the class wrapper here? Can't you just declare a constant? Let me know if this is a C# thing I don't understand.

exec/csnex/csnex.cs
```diff
@@ -0,0 +1,88 @@
...
+
+        private static Boolean ParseOptions(string[] args)
+        {
+            Boolean Retval = false;
+            for (int nIndex = 0; nIndex < args.Length; nIndex++) {
+                if (args[nIndex][0] == '-') {
+                    if (args[nIndex][1] == 'h' || args[nIndex][1] == '?' || ((args[nIndex][1] == '-' && args[nIndex][2] != '\0') && (args[nIndex][2] == 'h'))) {
+                        ShowUsage();
+                        Environment.Exit(1);
+                    } else if(args[nIndex][1] == 'n') {
```
ghewgill, line 34: Inconsistent spacing after `if`.

exec/csnex/csnex.cs
```diff
@@ -0,0 +1,88 @@
...
+                    } else {
+                        Console.Error.WriteLine(string.Format("Unknown option {0}\n", args[nIndex]));
+                        return false;
+                    }
+                } else {
+                    Retval = true;
+                    gOptions.Filename = args[nIndex];
+                }
+            }
+            if(gOptions.Filename.Length == 0) {
```
ghewgill, line 45: Inconsistent spacing after `if`.

exec/csnex/Exec.cs
```diff
@@ -0,0 +1,686 @@
...
+        {
+            throw new NotImplementedException(MethodBase.GetCurrentMethod().Name);
+        }
+#endregion
+#endregion
+
+        private int Loop()
+        {
+            while (ip < bytecode.codelen && exit_code == 0) {
+                switch (bytecode.code[ip]) {
```
ghewgill, line 578: Can you cast the byte here to `Opcode` so you don't have to have a `(byte)` cast in front of each case?

ghewgill (changes requested):

exec/csnex/Bytecode.cs
gitlarryf: Changed to local variable.

exec/csnex/Bytecode.cs
gitlarryf: Removed.

exec/csnex/Bytecode.cs
gitlarryf: Made local variable,

exec/csnex/CMakeLists.txt
gitlarryf: Fixed.

exec/csnex/Extensions.cs
gitlarryf: Opps.  Fixed.

exec/csnex/Extensions.cs
gitlarryf: Removed Extension.cs from project and use Array.Copy() static function now.

exec/csnex/CMakeLists.txt
gitlarryf: Fixed.

exec/csnex/Bytecode.cs
gitlarryf: Removed variables that do not need to be class members.

exec/csnex/Bytecode.cs
gitlarryf: Removed.

exec/csnex/Bytecode.cs
gitlarryf: Removed.

exec/csnex/Extensions.cs
gitlarryf: Removed entirely.

exec/csnex/Opcode.cs
gitlarryf: Yea.  Namespaces cannot contain fields or methods.  They have to be in a class.

exec/csnex/Exec.cs
gitlarryf: Fixed.

exec/csnex/csnex.cs
gitlarryf: Fixed.

exec/csnex/csnex.cs
gitlarryf: Fixed.

exec/csnex/Opcode.cs
gitlarryf: Moved BYTECODE_VERSION to Bytecode.cs instead.

exec/csnex/Bytecode.cs
```diff
@@ -0,0 +1,140 @@
...
+
+            /* Globals */
+            Get_VInt(obj, ref i);
+
+            /* String Table */
+            uint strtablesize = Get_VInt(obj, ref i);
+            strtable = GetStrTable(obj, strtablesize + i, ref i);
+
+            /* Types */
+            Get_VInt(obj, ref i);
```
ghewgill, line 87: It would be helpful, at least in development, to have an assertion here that the integer read from the bytecode is in fact 0. If it's not, you can detect an unimplemented bytecode section immediately rather than some other failure later.

ghewgill (changes requested):

exec/csnex/Bytecode.cs
gitlarryf: Added to all of the `Get_Vint()`'s.

PR #339, opened 2021-01-19
Implement boolean tests

gitlarryf:


exec/csnex/Exec.cs
```diff
@@ -43,7 +43,9 @@ public int Run(bool EnableAssertions)
 #region PUSHx Opcodes
         void PUSHB()
         {
-            throw new NotImplementedException(MethodBase.GetCurrentMethod().Name);
+            bool val = bytecode.code[ip + 1] != 0;
```
ghewgill, line 46: `bool` is used here and `Boolean` is used in other places. I think they're aliases for the same type in C#, but try to be consistent.

ghewgill (changes requested):

exec/csnex/Exec.cs
gitlarryf: Fixed.

PR #340, opened 2021-01-20
Implement JUMPTBL opcode and test

gitlarryf:


exec/csnex/Exec.cs
```diff
@@ -536,7 +539,19 @@ void JFCHAIN()
 
         void JUMPTBL()
         {
-            throw new NotImplementedException(MethodBase.GetCurrentMethod().Name);
+            ip++;
+            int val = Bytecode.Get_VInt(bytecode.code, ref ip);
+            Number n = stack.Pop().Number;
+            if (n.IsInteger() && !n.IsNegative()) {
+                int i = n;
+                if(i < val) {
```
ghewgill, line 547: Inconsistent spacing after `if`

exec/csnex/Number.cs
```diff
@@ -23,9 +23,34 @@ public static Number FromString(string str)
...
+#region Static Arithmetic Functions
+        public static Number Subtract(Number x, Number y)
+        {
+            return new Number(x.val - y.val);
+        }
+#endregion
+
+        public bool IsInteger()
+        {
+            Decimal i = Decimal.Round(val, MidpointRounding.AwayFromZero);
```
ghewgill, line 35: Why do this weird rounding mode? Truncating the number would be more appropriate here.

exec/csnex/Number.cs
```diff
@@ -23,9 +23,34 @@ public static Number FromString(string str)
...
+            return Decimal.Compare(val, i) == 0;
+        }
+
+        public bool IsNegative()
+        {
+            return Math.Sign(val) == -1;
+        }
+
+#region Overrides and Operators
+        public static implicit operator int(Number n)
```
ghewgill, line 45: Do you really want this to be an implicit automatic conversion? It seems like an implicit conversion would be more likely to introduce accidental hidden errors because you can't obviously tell when it is happening (by looking at the code). Suggest making this an explicit method call, instead of an implicit operator. Also, a 32 bit result is insufficient.

ghewgill (changes requested):

exec/csnex/Exec.cs
gitlarryf: Opps.  I think my Visual Studio formatter is still not quite perfect!  -- Fixed.

exec/csnex/Number.cs
gitlarryf: Changed to truncate `val`.

PR #342, opened 2021-01-21
Add better information to exceptions raised during runtime

gitlarryf:


exec/csnex/csnex.cs
```diff
@@ -76,12 +76,16 @@ static int Main(string[] args)
...
-            retval = exec.Run(gOptions.EnableAssertions);
+            try {
+                Bytecode bc = new Bytecode();
+                bc.LoadBytecode(gOptions.Filename, code);
 
+                Executor exec = new Executor(bc);
+                retval = exec.Run(gOptions.EnableAssertions);
+            } catch (Exception ex) {
+                Console.Error.WriteLine(ex.Message);
+                return -1;
```
ghewgill, line 87: -1 is not a valid process exit code. Use 1.

ghewgill (changes requested):

exec/csnex/csnex.cs
gitlarryf: Fixed.

PR #345, opened 2021-01-22
Add arithmetic functions for basic arithmetic test

gitlarryf:


exec/csnex/Number.cs
```diff
@@ -34,6 +34,38 @@ public static Number Subtract(Number x, Number y)
...
+                while (iy != 0) {
+                    if ((iy & 1) == 1) {
+                        r = Multiply(r, x);
+                    }
+                    x = Multiply(x, x);
+                    iy >>= 1;
+                }
+                return r;
+            }
+            return new Number(Decimal.ToInt64(x.val) ^ Decimal.ToInt64(y.val));
```
ghewgill, line 61: Just FYI at some point you'll need to handle noninteger exponents. But this is fine for now if it makes some tests pass.

PR #346, opened 2021-01-25
Add function support

gitlarryf:


exec/csnex/Exec.cs
```diff
@@ -11,27 +11,32 @@ public Executor(Bytecode bc)
...
+            Callstack = new Stack<int>();
             global = new Global(this);
             bytecode = bc;
+            ParamRecursionLimit = 2000; // ToDo: Add runtime$setRecursionLimit()
         }
 
         private int exit_code;
         private readonly Bytecode bytecode;
         public Stack<Cell> stack;
+        private Stack<int> Callstack;
```
ghewgill, line 23: The capitalization style here seems inconsistent. Should this be `callstack`?

exec/csnex/Exec.cs
```diff
@@ -11,27 +11,32 @@ public Executor(Bytecode bc)
...
         public bool enable_assert;
         private int ip;
         private Global global;
+        private Int32 ParamRecursionLimit;
```
ghewgill, line 27: Also inconsistent capitalization style here.

ghewgill (changes requested):

exec/csnex/Exec.cs
gitlarryf: Fixed.

exec/csnex/Exec.cs
gitlarryf: Fixed.

PR #349, opened 2021-02-04
Implement Exception handling in csnex.

gitlarryf:


exec/csnex/Exec.cs
```diff
@@ -278,7 +322,8 @@ void DIVN()
             Number b = stack.Pop().Number;
             Number a = stack.Pop().Number;
             if (b.IsZero()) {
-                throw new NeonDivideByZeroException("DivideByZeroException", "");
```
ghewgill, line 281: If the class `NeonDivideByZeroException` is no longer used, is it still defined somewhere? Shouldn't that declaration be removed now?

ghewgill (changes requested):

exec/csnex/Exec.cs
gitlarryf: Removed unused exceptions.

PR #357, opened 2021-02-19
Add sys module to csnex

gitlarryf:


exec/csnex/exclude.txt
```diff
@@ -240,7 +240,6 @@ string-test.neon
 strings.neon                                    # string__index
 struct-test.neon
 structure.neon
-sys-exit.neon                                   # sys module
```
ghewgill, line 243: Can you also enable the `cmdline.neon` test?

gitlarryf: Yes.  It's enabled now.

PR #359, opened 2021-02-21
Add Bytes support to csnex

gitlarryf:


exec/csnex/Extensions.cs
```diff
@@ -0,0 +1,36 @@
+﻿namespace csnex
+{
+    internal static partial class CodeExtensions
+    {
+        public static int Compare(this byte[] self, byte[] rhs)
```
ghewgill, line 5: Why is this function written this way, as a method? Shouldn't it just be an ordinary function that takes two `byte[]` parameters and returns an `int` result? Not everything needs to be a method.

exec/csnex/Extensions.cs
```diff
@@ -0,0 +1,36 @@
...
+                return 1;
+            } else if (self == null && rhs != null) {
+                if (rhs.Length == 0) {
+                    return 0;
+                }
+                return -1;
+            }
+
+            int res = 0;
+            for (int i = 0; i < self.Length && i < rhs.Length; i++) {
```
ghewgill, line 22: I don't think this works correctly as written. Compare two strings where one is a prefix of the other, like `car` and `carpet`. What should the result of this function be?

ghewgill (changes requested):

exec/csnex/Extensions.cs
gitlarryf: Rewrote to support better logic.

exec/csnex/Extensions.cs
```diff
@@ -0,0 +1,40 @@
...
+            for (i = 0; i < self.Length && i < rhs.Length; i++) {
+                if (self[i] > rhs[i]) {
+                    return 1;
+                } else if (self[i] < rhs[i]) {
+                    return -1;
+                }
+            }
+            if (i == self.Length && i == rhs.Length) {
+                return 0; // The lhs and rhs are proven to be equal to one another.
+            } else if (i  == self.Length) {
```
ghewgill, line 31: Fix spacing please.

ghewgill (changes requested):

exec/csnex/Extensions.cs
gitlarryf: Opps, sorry.  Good catch.  Fixed it.

PR #360, opened 2021-02-22
Exec csnex dictionary

gitlarryf:


exec/csnex/Exec.cs
```diff
@@ -652,12 +657,22 @@ void INDEXAN()
 
         void INDEXDR()
         {
-            throw new NotImplementedException(string.Format("{0} not implemented.", MethodBase.GetCurrentMethod().Name));
+            ip++;
+            string index = stack.Pop().String;
+            Cell addr = stack.Pop().Address;
+            if (addr.Dictionary.ContainsKey(index) == false) {
```
ghewgill, line 663: The `== false` here seems awkward. Use the `!` operator instead.

ghewgill (changes requested):

exec/csnex/Exec.cs
gitlarryf: Ok.  I wasn't sure when I wrote it that it sounded right; but since ContainsKey() returned a boolean, I went with the long-hand version.  I'll fix it.

PR #366, opened 2021-03-03
Exec csnex cleanup

gitlarryf:


exec/csnex/Bytecode.cs
```diff
@@ -1,12 +1,13 @@
...
 using System.Collections.Generic;
 using System.Diagnostics;
+using System.Reflection;
 
 namespace csnex
 {
     public class Bytecode
     {
-        private const int BYTECODE_VERSION = 3;
+        public static int BYTECODE_VERSION = Assembly.GetExecutingAssembly().GetName().Version.Revision;
```
ghewgill, line 10: I thought you were kidding when you said you were going to do this. Nope, make the bytecode version a literal constant like it was.

ghewgill (changes requested):
Please also fix spelling error in commit message for 15e1665.

exec/csnex/Bytecode.cs
gitlarryf: I don't disagree.  I thought I'd try it this way, so that if there was ever any question, you could easily look at the csnex version and know what version of the bytecode it can execute.  Too bad I can't do the opposite here either, and use the `BYTECODE_VERSION` in the `AssemblyVersion`

exec/csnex/csnex.cs
```diff
@@ -1,6 +1,14 @@
 using System;
 using System.Diagnostics;
+using System.Reflection;
 
+// Fixes MSBuild warning CA1016: Microsoft.Design : Add an AssemblyVersion attribute to 'csnex.exe'.
+// C# Versioning is as follows: <major version>.<minor version>.<build number>.<revision>
+// Major:     0 -- Version of csnex
+// Minor:    50 -- Minor version of csnex
+// Build:     0 -- Build # of csnex
+// Revision:  3 -- Version of Neon Bytecode this executor can execute.
```
ghewgill, line 10: I don't think it's a good idea to include this here. The bytecode version does not need to be part of the *software* version.

ghewgill (changes requested):

PR #368, opened 2021-03-08
Fixed csnex Cell get accessors

gitlarryf:


exec/csnex/Cell.cs
```diff
@@ -110,6 +143,9 @@ public enum Type
 
         public String String {
             get {
+                if (type == Type.None) {
+                    type = Type.String;
+                }
```
ghewgill, line 148: Does there need to be a `m_String = new String();` or something here to create a new one? Like you do for bytes, arrays, and dictionaries?

gitlarryf: I'm not sure I need to do that, since there can be such a thing as a null String.  But, I guess in reality, it couldn't hurt, and maybe it should in fact be there.

PR #370, opened 2021-03-10
Add Frames to csnex

gitlarryf:


exec/csnex/Exec.cs
```diff
@@ -7,6 +7,27 @@ namespace csnex
 {
     public class Executor
     {
+        sealed class ActivationFrame
+        {
+            public List<Cell> Locals;
+            public readonly int FrameSize;
```
ghewgill, line 13: Do you need to keep track of this separately? Isn't this exactly the same as the size of `Locals`?

exec/csnex/Exec.cs
```diff
@@ -7,6 +7,27 @@ namespace csnex
...
+            public readonly int NestingDepth;
+            public readonly ActivationFrame outer;
```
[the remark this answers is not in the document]
gitlarryf: Renamed.

exec/csnex/Exec.cs
gitlarryf: Removed FrameSize member.

exec/csnex/Exec.cs
```diff
@@ -908,10 +945,41 @@ void RESETC()
...
 #endregion
 #endregion
+        void dump_frames()
+        {
+            #if DUMP_FRAMES
+            Console.Out.Write("Frames:\n");
+            for (int i = 0; i < frames.Count; i++) {
+                ActivationFrame f = frames[i];
+                Console.Out.Write("\t {04:0}\t{ locals={04:1} }\n", i, f.FrameSize);
+                for (int l = 0; 0 < f.FrameSize; l++) {
```
ghewgill, line 955: You seem to have removed the `FrameSize` member but did not update this code. (This is why conditional compilation is annoying, because the compiler can't check your code.)

ghewgill (changes requested):

exec/csnex/Exec.cs
gitlarryf: Opps.  Yea, I don't care too much for the conditional complication either, but that's about the best option in C#.  I fixed this up so it is "current".

ghewgill: Conditional compilation is not the answer here. Use a constant (static) boolean value, and do `if (DUMP_FRAMES) ...`. Then when `DUMP_FRAMES` is `false`, the compiler will compile the code inside but won't emit any instructions for it. That is a far better option than `#if`.

ghewgill (changes requested):
As per comments around conditional compilation.

PR #371, opened 2021-03-11
Fix Cell copy semantics

gitlarryf:


exec/csnex/Cell.cs
```diff
@@ -309,6 +313,59 @@ public Boolean Equals(Cell x)
             }
             return false;
         }
+
+        public static void CopyCell(Cell dest, Cell source)
+        {
+            Debug.Assert(source != null);
+            Debug.Assert(dest != null);
```
ghewgill, line 320: Should you also assert that the source type is the same as the destination type? What happens if it isn't?

exec/csnex/Cell.cs
```diff
@@ -309,6 +313,59 @@ public Boolean Equals(Cell x)
...
+                    }
+                    break;
+                case Type.Number:
+                    dest.Number = new Number(source.Number);
+                    break;
+                case Type.Object:
+                    dest.m_Object = source.Object;
+                    break;
+                case Type.String:
+                    dest.String = String.Copy(source.String);
```
ghewgill, line 353: You don't need to make a copy here because C# `String` [is immutable](https://stackoverflow.com/questions/93091/why-cant-strings-be-mutable-in-java-and-net).

ghewgill (changes requested):

exec/csnex/Cell.cs
gitlarryf: No.  Since most new cells start off life as a `Type.None`, I want them to be able to morph here.  If they are different types, then it will literally copy the source over the destination cell blindly, which is ok since it also types the type member.

exec/csnex/Cell.cs
gitlarryf: Fixed.

exec/csnex/Cell.cs
ghewgill: I must be missing it, where is the type member copied?

PR #373, opened 2021-03-12
Add remaining global array functions to csnex

gitlarryf:


exec/csnex/Extensions.cs
```diff
@@ -47,5 +47,51 @@ public static List<Cell> ToCellArray(this List<string> self)
...
+                if (!self[i].Equals(rhs[i])) {
+                    return false;
+                }
+            }
+            return true;
+        }
+
+        public static string Quote(this string s)
+        {
+            string r = "\"";
```
ghewgill, line 66: Since `string` is immutable, for performance here you will want to use `StringBuilder`.

exec/csnex/Global.cs
```diff
@@ -68,13 +68,195 @@ public void array__extend()
...
+            Number index = Exec.stack.Pop().Number;
+            List<Cell> array = Exec.stack.Pop().Address.Array;
+
+            if (!index.IsInteger()) {
+                Exec.Raise("ArrayIndexException", index.ToString());
+                return;
+            }
+
+            int i = Number.number_to_int32(index);
+            array[i].ResetCell();
```
ghewgill, line 129: You probably need some range checking here.

exec/csnex/Global.cs
```diff
@@ -68,13 +68,195 @@ public void array__extend()
...
+            Number new_size = Exec.stack.Pop().Number;
+            List<Cell> array = Exec.stack.Pop().Address.Array;
+
+            if (!new_size.IsInteger()) {
+                Exec.Raise("ArrayIndexException", new_size.ToString());
+                return;
+            }
+
+            if (Number.number_to_int32(new_size) < array.Count) {
+                array.RemoveRange(Number.number_to_int32(new_size), array.Count - Number.number_to_int32(new_size));
```
ghewgill, line 144: This seems inconsistent. When you remove a single element (in the previous function), you call `.ResetCell()`, but you don't do that here. Which is correct?

ghewgill (changes requested):

exec/csnex/Extensions.cs
gitlarryf: Good call.  Set this to be a `StringBuilder()`.

exec/csnex/Global.cs
gitlarryf: I suppose it's not needed at all.  I removed it from the previous call.

exec/csnex/Global.cs
gitlarryf: Checking against what?  The compiler would never generate code that would violate this, would it?

ghewgill: It's not about code generation. What if `i` is out of range of `array`?

gitlarryf: Made range checking match global.cpp.

gitlarryf: > It's not about code generation. What if `i` is out of range of `array`?

Yea.  I considered that, and so I copied the range checking out of global.cpp.

PR #374, opened 2021-03-17
Add io module to csnex

gitlarryf:


exec/csnex/lib/io.cs
```diff
@@ -0,0 +1,210 @@
...
+            switch ((Mode)Number.number_to_uint32(mode)) {
+                case Mode.read:
+                    fa = FileAccess.Read;
+                    fm = FileMode.Open;
+                    break;
+                case Mode.write:
+                    fa = FileAccess.Write;
+                    fm = FileMode.Append;
+                    break;
+            default:
```
ghewgill, line 85: Indenting here should match the cases.

ghewgill (changes requested):

exec/csnex/lib/io.cs
gitlarryf: Opps.  Seems no matter how you configure your IDEs formatting settings, it still tries really hard to format the code the way _it_ wants to format it.  :/  Fixed it.

PR #375, opened 2021-03-18
Add Neon runtime information to NeonException

gitlarryf:


exec/csnex/Exceptions.cs
```diff
@@ -16,6 +23,21 @@ public class NeonException: ApplicationException
...
+
+        // Satisfy Warning CA2240 to implement a GetObjectData() to our custom exception type.
+        [SecurityPermission(SecurityAction.LinkDemand, Flags = SecurityPermissionFlag.SerializationFormatter)]
+        public override void GetObjectData(SerializationInfo info, StreamingContext context)
+        {
+            if (info == null) {
+                throw new ArgumentNullException("info");
+            }
+            info.AddValue("NeonException", Name);
+            info.AddValue("NeonInfo", info);
```
ghewgill, line 35: Is this right? You seem to be adding `info` to itself, which is likely to end badly. Is this supposed to add `Info` instead?

ghewgill (changes requested):

exec/csnex/Exceptions.cs
gitlarryf: Ah.  Yes, this is wrong.  Fixed.

PR #376, opened 2021-03-20
Add Object support to csnex

gitlarryf:


exec/csnex/Global.cs
```diff
@@ -464,11 +484,209 @@ public void number__toString()
         }
 #endregion
 #region Object Functions
+        #region Object Getters
+        public void object__getArray()
+        {
+            Object obj = Exec.stack.Pop().Object;
+
+            List<Object> a = new List<Object>();
```
ghewgill, line 492: There's no purpose in creating a new `List` here, only to throw it away in the next call. Either don't initialize this reference (if the compiler lets you) or initialize it to null.

exec/csnex/Global.cs
```diff
@@ -464,11 +484,209 @@ public void number__toString()
...
+            }
+
+            Exec.stack.Push(new Cell(r));
+        }
+
+        public void object__getNumber()
+        {
+            Cell obj = Exec.stack.Pop();
+
+            Number n = new Number(0);
```
ghewgill, line 569: No need to create a new number here.

exec/csnex/Object.cs
```diff
@@ -15,30 +49,210 @@ public virtual string toLiteralString()
...
+        }
+
+        public override bool subscript(Object index, ref Object r)
+        {
+            Number i = new Number(0);
+            if (!index.getNumber(ref i)) {
+                throw new NeonRuntimeException("DynamicConversionException", "to Number");
+            }
+            int ii = Number.number_to_int32(i);
+            if (ii >= Value.Count) {
```
ghewgill, line 72: What happens if `ii` is negative here?

exec/csnex/Object.cs
```diff
@@ -15,30 +49,210 @@ public virtual string toLiteralString()
...
+        {
+            StringBuilder r = new StringBuilder("{");
+            bool first = true;
+            foreach (KeyValuePair<string, Object> e in Value) {
+                if (first) {
+                    first = false;
+                } else {
+                    r.Append(", ");
+                }
+                r.Append("\"");
```
ghewgill, line 199: Don't manually quote this here, use your `toLiteralString` or whatever. What if the key value itself contains quotes?

ghewgill (changes requested):

exec/csnex/Global.cs
gitlarryf: Fixed.

exec/csnex/Global.cs
gitlarryf: Fixed.

exec/csnex/Object.cs
gitlarryf: Added a test for negative values.

exec/csnex/Object.cs
gitlarryf: Calling `string.Quote()` now instead.

PR #382, opened 2021-04-06
Add module support to csnex

gitlarryf:


exec/csnex/Support.cs
```diff
@@ -0,0 +1,84 @@
...
+{
+    public class Support
+    {
+        static public List<string> Paths;
+
+        public Support(string source_path)
+        {
+            Paths = new List<string>();
+            // Always check current directory;
+            Paths.Add(string.Format(".{0}", Path.DirectorySeparatorChar));
```
ghewgill, line 15: This seems significantly different from the code in support.cpp. The order of paths should be the same as what is already established. Your code changes the order here with no information about why.

exec/csnex/Support.cs
```diff
@@ -0,0 +1,84 @@
...
+            // Also check the Neon executable path.
+            Paths.Add(string.Format("{0}{1}", source_path, Path.DirectorySeparatorChar));
+            // Then check the cwd.
+            Paths.Add(Environment.CurrentDirectory);
```
ghewgill, line 19: What is the purpose of this? You have already added the current directory above.

exec/csnex/Support.cs
```diff
@@ -0,0 +1,84 @@
...
+                string[] evpaths = ne.Split(Path.PathSeparator);
+                foreach (string s in evpaths) {
+                    // Don't bother pushing empty paths.
+                    if (s != null && s.Length > 0) {
+                        Paths.Add(s);
+                    }
+                }
+            }
+            // Next, check for .neonpath, and process any paths that might exist in it.
+            if (File.Exists(".neonpath")) {
```
ghewgill, line 33: This sort of existence check is not robust code, it is a potential race condition. It's possible that the file could be removed by something else between the time you check the existence and try to open it. Instead, simply try to open the file and handle any failure to open in the usual way.

exec/csnex/Support.cs
```diff
@@ -0,0 +1,84 @@
...
+                StreamReader sr = new StreamReader(".neonpath");
+                while (!sr.EndOfStream) {
```
ghewgill, line 35: This kind of "end of file" checking is also not robust. The `ReadLine` function already tells you when you are at the end of the file (it returns `null`). Use that instead.

exec/csnex/Support.cs
```diff
@@ -0,0 +1,84 @@
...
+            // Then check the cwd.
+            Paths.Add(Environment.CurrentDirectory);
+
+            // Get paths from %NEONPATH% environment variable, if there is one.
+            string ne = Environment.GetEnvironmentVariable("NEONPATH");
+            if (ne != null) {
+                string[] evpaths = ne.Split(Path.PathSeparator);
+                foreach (string s in evpaths) {
+                    // Don't bother pushing empty paths.
+                    if (s != null && s.Length > 0) {
```
ghewgill, line 27: Under what conditions would `Split()` return a `null` in the array? I don't see that mentioned anywhere in the documentation. You do not need to handle the `null` case if it is not documented to happen.

ghewgill (changes requested):

exec/csnex/Exec.cs
```diff
@@ -26,13 +27,29 @@ public ActivationFrame(int nest, ActivationFrame outer, int size, int stack_dept
             }
         }
 
-        public Executor(Bytecode bc)
+        sealed class StackFrame
```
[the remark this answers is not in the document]
gitlarryf: Renamed to CallStack, to match cnex naming convention.

exec/csnex/Exec.cs
```diff
@@ -791,17 +841,38 @@ void CALLP()
...
+            int mod = Bytecode.Get_VInt(module.Bytecode.code, ref ip);
+            int fun = Bytecode.Get_VInt(module.Bytecode.code, ref ip);
+            int efi = 0;
+            if (callstack.Count >= param_recursion_limit) {
+                Raise("StackOverflowException", "");
+                return;
+            }
+
+            Module m = modules[module.Bytecode.strtable[mod]];
+            //Module m = LocateModule(module.Bytecode.strtable[mod]);
```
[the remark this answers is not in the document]
gitlarryf: Removed.

exec/csnex/Support.cs
gitlarryf: Changed order of the load paths.

exec/csnex/Support.cs
gitlarryf: Removed.

exec/csnex/Support.cs
gitlarryf: Use a try/catch for this now.

exec/csnex/Support.cs
gitlarryf: Removed sr.EndOfStream check.

exec/csnex/Support.cs
gitlarryf: Removed != null check.

ghewgill: The check still seems to be there?

exec/csnex/Support.cs
```diff
@@ -0,0 +1,87 @@
...
+                }
+            }
+            // Next, check for .neonpath, and process any paths that might exist in it.
+            try {
+                StreamReader sr = new StreamReader(".neonpath");
+                string line = null;
+                do {
+                    line = sr.ReadLine();
+                    // Don't bother pushing empty paths.
+                    if (line != null && line.Length > 0) {
```
ghewgill, line 40: You've got confusing logic here and a double check for null. Check once for null and use break to get out of the loop.

exec/csnex/Support.cs
```diff
@@ -0,0 +1,87 @@
...
+                        Paths.Add(line);
+                    }
+                } while (line != null);
+            } catch {
```
ghewgill, line 44: It is inappropriate to catch and ignore *all* exceptions.

ghewgill (changes requested):

PR #386, opened 2021-04-21
Add class, interface and pointer support to csnex

gitlarryf:


exec/csnex/Cell.cs
```diff
@@ -276,13 +303,20 @@ public static string toString(Cell c)
         public override String ToString()
         {
             switch (type) {
+                case Type.Address:
+                    if (Address != null) {
+                        return Address.ToString();
+                    }
+                    return "null";
```
ghewgill, line 310: Use `NIL`.

exec/csnex/Cell.cs
```diff
@@ -300,13 +334,15 @@ public Boolean Equals(Cell x)
...
+                    // Not implemented yet.
                     return false;
                 case Type.Number:
                     return Number.IsEqual(Number, x.Number);
                 case Type.Object:
-                    // Not Implemented yet.
+                    // Not implemented yet.
                     return false;
+                case Type.Other:
+                    return Other.Equals(x.Other);
```
ghewgill, line 345: I think you want to use `==` here, rather than `.Equals`. You don't want to do a deep value comparison, you only need to check whether the references are the same.

exec/csnex/Cell.cs
```diff
@@ -431,6 +470,15 @@ public void Set(Object o)
             Object = o;
         }
 
+        public void Set(object o)
```
ghewgill, line 473: I'm not sure it's wise to have an overload for the generic C# `object`. Doing so means that any call to `Set` where you pass anything that doesn't match any other overload, ends up here even if you didn't mean to. I would prefer to see `SetObject` here (and corresponding named methods for other cell types). There is no need for overloading here.

exec/csnex/Cell.cs
```diff
@@ -165,16 +181,20 @@ public enum Type
         }
 
         public Type type { get; private set; }
+
+        public UInt64 AllocNum { get; private set; }
```
ghewgill, line 185: Why does this declare a setter? Can you just say `{ get; }` because that's all you need?

exec/csnex/Exec.cs
```diff
@@ -257,7 +259,8 @@ void PUSHI()
 
         void PUSHNIL()
         {
-            throw new NotImplementedException(string.Format("{0} not implemented.", MethodBase.GetCurrentMethod().Name));
+            ip++;
+            stack.Push(new Cell());
```
ghewgill, line 263: The `NIL` value is actually a pointer value and you should probably use `Cell(null)` instead. There's a difference between pushing a cell that doesn't contain anything, versus pushing a cell that contains a pointer which happens to be `NIL`.

ghewgill (changes requested):

exec/csnex/Cell.cs
gitlarryf: Fixed.

exec/csnex/Cell.cs
gitlarryf: Changed to `m_Other == x.m_Other`

exec/csnex/Cell.cs
gitlarryf: The next pull request will revamp the `csnex.Cell` class.

exec/csnex/Cell.cs
gitlarryf: Ah.  Good call.  I assumed I needed one for the constructor to be able to set it, forgetting that the constructor isn't accessing it as a property, but rather a local variable.  Removed the `set;`.

exec/csnex/Exec.cs
gitlarryf: Changed to `new Cell((object)null)` to avoid ambiguous call between `Cell.Cell(List<Cell>)`, and `Cell.Cell(byte[])`.  This will be cleaned up in the upcoming `csnex.Cell()` class clean up.

PR #388, opened 2021-04-30
Add trace output to csnex

gitlarryf:


exec/csnex/Exec.cs
```diff
@@ -50,8 +50,10 @@ public Executor(Module mod)
             init_order = new List<string>();
             module = mod;
             modules.Add("", mod);
+            Disassembler = new InstructionDisassembler(mod.Bytecode, ip);
             param_recursion_limit = 1000;
             Allocations = 0;
+            EnableTrace = csnex.gOptions.EnableTracing;
```
ghewgill, line 56: What is the purpose of making a class member copy of this flag? Why not refer to `csnex.gOptions.EnableTracing` directly when needed?

exec/csnex/Exec.cs
```diff
@@ -1165,6 +1168,9 @@ private void Invoke(Module m, int index)
         private int Loop(Int64 min_callstack_depth)
         {
             while (callstack.Count > min_callstack_depth && ip < module.Bytecode.code.Length && exit_code == 0) {
+                if (EnableTrace) {
+                    Console.Error.Write("mod {0} ip {1} ({2}) {3}\n", module.Name, ip, stack.Count, Disassembler.DisassembleInstruction(module.Bytecode, ip));
```
ghewgill, line 1172: Would it make more sense to use `WriteLine` here and omit the embedded `\n`?

ghewgill (changes requested):

exec/csnex/Exec.cs
gitlarryf: Fixed.

exec/csnex/Exec.cs
gitlarryf: I wanted to make `Executor` self contained.  I fixed this by passing `csnex.gOptions.EnableTracing` as a constructor parameter.

exec/csnex/Disassembly.cs
```diff
@@ -0,0 +1,790 @@
...
+
+        void INDEXAN()
+        {
+            stream.AppendFormat("INDEXAN");
+            index++;
+        }
+
+        void INDEXDR()
+        {
+            stream.AppendFormat("INDEXAN");
```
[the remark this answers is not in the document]
gitlarryf: Fixed.

exec/csnex/Disassembly.cs
```diff
@@ -0,0 +1,790 @@
...
+                case Opcode.CALLV: CALLV(); break;
+                case Opcode.PUSHCI: PUSHCI(); break;
+                default:
+                    stream.AppendFormat("Unknown opcode: {0}", bytecode.code[index]);
+                    index++;
+                    break;
+            }
+        }
+
+        public string DisassembleInstruction(Bytecode obj, int index)
```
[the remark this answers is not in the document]
gitlarryf: Fixed by deriving the Executor from the Disassembler class.  This fixes the whole problem for me.  In fact, I might even be able to clean up the code a little more by not even passing the `bytecode` and `ip`, but rather just assigning them in the `base` class, and referencing them from there.

exec/csnex/Disassembly.cs
```diff
@@ -0,0 +1,788 @@
...
+                case Opcode.CALLV: CALLV(); break;
+                case Opcode.PUSHCI: PUSHCI(); break;
+                default:
+                    stream.AppendFormat("Unknown opcode: {0}", bytecode.code[index]);
+                    index++;
+                    break;
+            }
+        }
+
+        public string DisassembleInstruction(Bytecode obj, int ip)
```
ghewgill, line 779: There's still no reason for this function to be a member function. Make it a `static` function, and construct a new `InstructionDisassembler` here to do the actual work.

ghewgill (changes requested):

PR #391, opened 2021-05-15
Fix all toString() implementations in csnex

gitlarryf:


exec/csnex/Object.cs
```diff
@@ -179,11 +179,12 @@ public override bool subscript(Object index, ref Object r)
 
             if (Value.ContainsKey(s)) {
                 r = Value[s];
+                return true;
             }
             if (r == null) {
```
ghewgill, line 184: I think this logic can be simplified, as I can no longer see a code path here where `r != null`.

ghewgill (changes requested):

exec/csnex/Object.cs
gitlarryf: Fixed.  Removed the `r==null` logic.

PR #395, opened 2021-05-20
Add opcode INDEXAN to csnex

gitlarryf:


exec/csnex/Exec.cs
```diff
@@ -813,7 +813,26 @@ void INDEXAV()
...
+            if (!index.IsInteger()) {
+                RaiseLiteral("ArrayIndexException", Cell.CreateStringCell(index.ToString()));
+                return;
+            }
+            int i = Number.number_to_int32(index);
+            if (i < 0) {
+                RaiseLiteral("ArrayIndexException", Cell.CreateStringCell(new Number(i).ToString()));
+                return;
+            }
+            if (i >= array.Array.Count) {
```
ghewgill, line 829: Probably better to combine these two tests (this and the one above) into an 'or' condition, since the exception raised is identical.

ghewgill (changes requested):

PR #400, opened 2021-06-15
Add console module to csnex

gitlarryf:


exec/csnex/lib/console.cs
```diff
@@ -0,0 +1,41 @@
+﻿using System;
+using System.IO;
+using System.Security.Cryptography;
```
ghewgill, line 3: What is this using statement for? Surely this module doesn't do any cryptography.

ghewgill (changes requested):

exec/csnex/lib/console.cs
gitlarryf: Fixed from online edit.

PR #406, opened 2021-08-27
Support built in method calls for known Objects in cnex

gitlarryf:


exec/cnex/global.c
```diff
@@ -1218,6 +1218,46 @@ void object__invokeMethod(TExecutor *exec)
...
 
+    if (obj->object->type == oArray) {
+        if (strcmp(name, "size") == 0) {
+            if (args->array != NULL && args->array->size != 0) {
+                exec->rtl_raise(exec, "DynamicConversionException", "invalid number of arguments to size() (expected 0)");
+                goto cleanup;
+            }
+            push(exec->stack, cell_fromObject(object_createNumberObject(number_from_uint64(((Cell*)obj->object->ptr)->array->size))));
+            goto cleanup;
+        }
```
ghewgill, line 1229: Need an `else` clause here to catch unknown method names and raise an exception.

exec/cnex/global.c
```diff
@@ -1218,6 +1218,46 @@ void object__invokeMethod(TExecutor *exec)
...
+                Cell *key = cell_fromObject(object_createStringObject(keys->array->data[i].string));
+                array_appendElement(okeys, key);
+                cell_freeCell(key);
+            }
+            Cell *oArray = cell_fromObject(object_createArrayObject(okeys));
+            push(exec->stack, oArray);
+            cell_freeCell(keys);
+            array_freeArray(okeys);
+            goto cleanup;
+        }
```
ghewgill, line 1258: Also handle unknown method names here.

exec/cnex/global.c
```diff
@@ -1218,6 +1218,46 @@ void object__invokeMethod(TExecutor *exec)
...
+                goto cleanup;
+            }
+            push(exec->stack, cell_fromObject(object_createNumberObject(number_from_sint64(((Cell*)obj->object->ptr)->dictionary->len))));
+            goto cleanup;
+        } else if (strcmp(name, "keys") == 0) {
+            if (args->array != NULL && args->array->size != 0) {
+                exec->rtl_raise(exec, "DynamicConversionException", "invalid number of arguments to keys() (expected 0)");
+                goto cleanup;
+            }
+            // ToDo: Consider doing an "inplace" conversion, where the "keys" Cell is manually converted into an Object Array.
```
ghewgill, line 1245: What does this comment mean? What would be the benefit of changing the way this works now?

ghewgill (changes requested):

exec/cnex/global.c
gitlarryf: That was a mental note to myself, to see if I could maybe keep the same Cell Array, but convert all the members to Object Strings, from TStrings, without having to build a whole new array.  I wanted to write the code to try it out, but then realized I was optimizing a problem I didn't even know I had; so, I left it as a comment, and walked away.

exec/cnex/global.c
gitlarryf: Fixed.

exec/cnex/global.c
gitlarryf: Fixed.

PR #407, opened 2021-08-31
Implement correct number exceptions in cnex

gitlarryf:


exec/cnex/cnex.c
```diff
@@ -480,6 +481,22 @@ void exec_rtl_raiseException(TExecutor *self, const char *name, const char *info
     string_freeString(i);
 }
 
+void exec_numberCheckAndRaise(TExecutor *self, unsigned int start_ip, const char *what)
+{
+    if (_IDEC_glbflags & BID_OVERFLOW_EXCEPTION) {
+        self->ip = start_ip;
+        exec_rtl_raiseException(self, "NumberException.Overflow", what);
```
ghewgill, line 488: Do you want to `return;` here in case more than one of these flags is set? Or use `else if`?

gitlarryf: I thought about it, but figured that it might be possible for more than one bit to be set, and figured I'd allow the code to basically go into a "first or worst" case.  Looking back?  Yea.  I probably should just `return` after any exception is raised.

gitlarryf: Added a `return;` after every possible exception.  Now there will be no fall-thru.

PR #419, opened 2022-06-20
Fixed memory leak in cnex dictionary__toString__string function

gitlarryf:


exec/cnex/nstring.c
```diff
@@ -525,6 +525,20 @@ TString *string_toUpperCase(TString *s)
     return r;
 }
 
+// Note: This function will append quotes to an existing string, then return it.
+// It is important to note that you do not want to call this function unless you
+// intend to modify the underlying string object!  This is useful for avoiding
+// a lot of string management in certain cases.
+TString *string_appendQuotes(TString *s)
```
ghewgill, line 532: This seems like a poor name for this function, because "append" means "add to the end of". But this puts quotes around the whole thing, start and end, right?

ghewgill (changes requested):

exec/cnex/nstring.c
gitlarryf: Fixed.  Change function to `string_quoteInPlace()`.

PR #425, opened 2022-08-09
Fix bug in cnex net_socket_recv() function

gitlarryf:


exec/cnex/lib/net.c
```diff
@@ -142,6 +142,7 @@ void net_socket_bind(TExecutor *exec)
     if (r < 0) {
         // ToDo: Raise exception here?
         perror("bind");
+        // Code will fall through to bail:, so a goto isn't necessary here.
```
ghewgill, line 145: Don't take shortcuts, add the goto. That way: (1) the code will look right because it looks the same as all the other error cases, and (2) you won't get bit when someday you add some code after this if block.

exec/cnex/lib/net.c
```diff
@@ -249,7 +251,9 @@ void net_socket_recv(TExecutor *exec)
     int r = recv(s, (char*)buf, n, 0);
     if (r < 0) {
         perror("recv");
+        // Should we push a FALSE on the stack here?
```
ghewgill, line 254: What does this comment mean? You either need to or you don't. In this case, you don't.

ghewgill (changes requested):

exec/cnex/lib/net.c

exec/cnex/lib/net.c
gitlarryf: Removed comment.

PR #426, opened 2022-08-10
Fix destroyStack() function in cnex

gitlarryf:


ghewgill:
As discussed previously, this would be better solved by changing the stack offsets so 0 represents an empty stack, not -1.

PR #431, opened 2022-10-13
Add array preallocation to cnex

gitlarryf:


exec/cnex/array.c
```diff
@@ -45,13 +53,17 @@
...
     }
 
-    self->data = realloc(self->data, sizeof(Cell) * (self->size + iElements));
-    if (self->data == NULL) {
-        fatal_error("Unable to allocate memory for %d expanded array elements.", self->size + iElements);
+    if (self->size + iElements > self->max) {
+        self->max = self->size + iElements;
+        self->data = realloc(self->data, sizeof(Cell) * self->max);
+        if (self->data == NULL) {
+            fatal_error("Unable to allocate memory for %d expanded array elements.", self->max);
```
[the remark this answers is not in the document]
gitlarryf: Changed format specifier to %zu.

exec/cnex/array.c
```diff
@@ -27,10 +32,13 @@
...
     a->size = iElements;
-    a->data = malloc(sizeof(Cell) * a->size);
+    if (iElements > a->max) {
+        a->max = iElements;
+        a->data = realloc(a->data, sizeof(Cell) * a->max);
+    }
     a->refcount = 1;
     if (a->data == NULL) {
-        fatal_error("Unable to allocate memory for %d array elements.", a->size);
+        fatal_error("Unable to allocate memory for %d array elements.", a->max);
```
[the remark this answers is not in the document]
gitlarryf: Changed format specifier to %zu.

exec/cnex/array.c
```diff
@@ -27,10 +32,13 @@ Array *array_createArrayFromSize(size_t iElements)
     Array *a = array_createArray();
 
     a->size = iElements;
-    a->data = malloc(sizeof(Cell) * a->size);
+    if (iElements > a->max) {
+        a->max = iElements;
+        a->data = realloc(a->data, sizeof(Cell) * a->max);
```
ghewgill, line 37: If you're going to have a function to create an array with a specific size, it doesn't make sense to allocate some and then immediately realloc it to what you actually wanted. Just allocate what you want, once.

exec/cnex/array.c
```diff
@@ -45,13 +53,17 @@ Array *array_expandArray(Array *self, size_t iElements)
         if (self == NULL) {
             fatal_error("Unable to allocate array for array expansion.");
         }
-        self->size = 0;
-        self->data = NULL;
+        self->size = iElements;
```
ghewgill, line 56: Under what conditions does "self" come through as NULL here? Do you need this code path? And if you do, then this block could be replaced with a call to `array_createArray()`.

exec/cnex/array.c
```diff
@@ -45,13 +53,17 @@ Array *array_expandArray(Array *self, size_t iElements)
...
+        self->max = ARRAY_INITIAL_SIZE;
+        self->data = malloc(sizeof(Cell) * self->max);
     }
 
-    self->data = realloc(self->data, sizeof(Cell) * (self->size + iElements));
-    if (self->data == NULL) {
-        fatal_error("Unable to allocate memory for %d expanded array elements.", self->size + iElements);
+    if (self->size + iElements > self->max) {
+        self->max = self->size + iElements;
```
ghewgill, line 62: Isn't this where you want to double the allocation size? If you were to call `array_expandArray()` one element at a time, you'd be back to where you started as far as reallocating and copying.

exec/cnex/array.c
```diff
@@ -83,9 +95,13 @@ void array_clearArray(Array *self)
         for (size_t i = 0; i < self->size; i++) {
             cell_clearCell(&self->data[i]);
         }
-        free(self->data);
-        self->data = NULL;
+        // ...Wait.  Why are we freeing the array here?  Why bother?  Just leave it, with self->max set, and
```
ghewgill, line 98: This seems to be a "note to self" comment that no longer needs to be present. The code it's referring to is not present.

exec/cnex/array.c
```diff
@@ -107,6 +123,7 @@ Array *array_copyArray(Array *a)
     for (size_t i = 0; i < r->size; i++) {
         cell_copyCell(&r->data[i], &a->data[i]);
     }
+    r->max = a->max;
```
ghewgill, line 126: This seems very wrong here. You've called `array_createArrayFromSize` with the original array size, but then you're saying the max is whatever the original array had as a max. But you haven't allocated space for max in this array.

exec/cnex/array.c
```diff
@@ -135,14 +152,17 @@ size_t array_appendElement(Array *self, const Cell *element)
...
+            self->data = realloc(self->data, sizeof(Cell) * self->max);
+            if (self->data == NULL) {
+                fatal_error("Not enough memory to expand array.");
+            }
         }
         self->size++;
     }
     if (self->data == NULL) {
-        self->data = malloc(sizeof(Cell));
+        self->data = malloc(sizeof(Cell) * self->max);
```
ghewgill, line 165: Under what conditions is `self->data` NULL, but `self->max` is nonzero? What does a nonzero max mean if there is no data? What if `self->max` is zero here? Line 169 seems to assume it's at least 1.

exec/cnex/cell.c
```diff
@@ -523,16 +523,24 @@ Cell *cell_arrayIndexForWrite(Cell *c, size_t i)
         // Assign the (new) copied array to the cell object.
         c->array = wa;
     }
-    if (i >= c->array->size) {
-        c->array->data = realloc(c->array->data, sizeof(Cell) * (i+1));
+    if (i >= c->array->max) {
+        // ToDo: Fix this!  All array memory management should happen in array.c.
```
ghewgill, line 527: This needs to be fixed before it can be accepted.

ghewgill (changes requested):

exec/cnex/array.c
gitlarryf: Changed operation of this function to allocate its own `Array`, and allocate (only once) enough room for the required elements.

exec/cnex/array.c
gitlarryf: I'm not positive I need it either.  I've left the code path, but called `array_createArray()` in case this code path is ever reached.

exec/cnex/array.c
gitlarryf: Removed.

exec/cnex/array.c
gitlarryf: Removed.  `max` is already set in this case, as you mentioned.

exec/cnex/array.c
gitlarryf: So, this was a unique scenario.  This happened in the startup routine of cnex, building the arguments, for one.  A call to `cell_createCellArray(0)` is made, then each arg is added to the array with a call to `cell_arrayAppendElement()`.  This was a case of `self->data` being NULL, as was `self->max`.  All of this code has been reworked based on the review comments so the above description will no longer happen.  Now, `self->data` should _never_ be NULL.

exec/cnex/cell.c
gitlarryf: Fixed.

exec/cnex/array.c
```diff
@@ -9,29 +9,41 @@
...
+    Array *a = malloc(sizeof(Array));
+    if (a == NULL) {
+        fatal_error("Unable to allocate array for size.");
+    }
 
+    a->data = NULL;
     a->size = iElements;
-    a->data = malloc(sizeof(Cell) * a->size);
-    a->refcount = 1;
+    a->max = (a->size > 0 ? a->size * 2 : ARRAY_INITIAL_SIZE);
```
ghewgill, line 40: You don't need to preallocate 2x the storage here. You're creating an array with a known number of elements, like maybe a copy of another array, or perhaps a literal constant. Allocate only what is required here, and then later IF you start adding new elements, then preallocate some more.

exec/cnex/array.c
```diff
@@ -50,17 +62,21 @@ Array *array_resizeArray(Array *self, size_t newSize)
             cell_clearCell(&self->data[i]);
         }
     }
-    if (newSize > 0) {
-        self->data = realloc(self->data, sizeof(Cell) * newSize);
+    if (newSize > 0 && newSize > self->max) {
+        self->max = newSize * 2;
```
ghewgill, line 66: Again, this seems excessive. If you create an empty array, and then resize it to 1000 elements, you don't want to actually allocate 2000. Instead, you can do 2*max if that's sufficient, or if the requested size is more than that, just allocate the requested size.

ghewgill (changes requested):

exec/cnex/array.c
gitlarryf: Ok.  That makes sense.  I will keep the check for `> 0` however, since some of the array functions start off with a zero-size array, and start appending elements, so I want to have space already allocated for elements.  But if we're called with fixed size, then correct; we're most likely performing a copy, or a CONSA.  I put this logic in originally for the module init code cnex handles on startup, mainly.

PR #433, opened 2022-10-19
Fix SQL bindings for variables containing underscores (Fixes #432)

### LINKED ISSUE #432: Embedded SQL does not allow variables with underscores.
The following embedded SQL code:
```
LET col_val := "Value"
EXEC SQL INSERT INTO table VALUES (:col_val);
```
Results in a compile error:

```
2| EXEC SQL INSERT INTO table VALUES (:col_val);
                                      ~~~~
                                      ^
Error N4305: 2:36 variable not found
```

gitlarryf:


t/sql-embed.neon
```diff
@@ -5,9 +5,9 @@ EXEC SQL CONNECT TO :DatabaseName;
 EXEC SQL DROP TABLE IF EXISTS user;
 EXEC SQL CREATE TABLE user(name STRING PRIMARY KEY, password STRING);
 
-LET name: String := "name"
+LET user_name: String := "name"
```
ghewgill, line 8: Don't change this existing test here. Instead, add a few new lines to the bottom of this file that tests the use of an underscore in a parameter value.

ghewgill (changes requested):

t/sql-embed.neon
gitlarryf: Added a new test below the current one.  I only used one parameter with an underscore; should I have used two or more?

ghewgill: One is sufficient.

ghewgill:
The commit message for this should reference the issue it fixes. (The title of the pull request, where you have mentioned the issue number, means nothing.)

PR #434, opened 2022-10-24
Fix select() function in cnex

gitlarryf:


exec/cnex/lib/net.c
```diff
@@ -355,7 +364,7 @@ void net_socket_select(TExecutor *exec)
     }
 
     for (size_t i = 0; i < read->array->size; ) {
-        SOCKET *ps = check_socket(exec, read->array->data[i].object);
+        SOCKET *ps = check_socket(exec, ((Array*)read->array->data[i].array)->data[0].object);
```
ghewgill, line 367: I am still pretty sure that you do not need these casts. If you don't need them, remove them.

ghewgill (changes requested):

exec/cnex/lib/net.c

PR #435, opened 2022-11-02
Improve array memory management in cnex.

gitlarryf:


exec/cnex/array.c
```diff
@@ -40,24 +40,58 @@ Array *array_createArrayFromSize(size_t iElements)
 
 Array *array_expandArray(Array *self, size_t iElements)
```
ghewgill, line 41: What's the difference between calling array_expandArray and calling array_resizeArray? Why not just have array_resizeArray which can make the array both bigger and smaller?

ghewgill (changes requested):

exec/cnex/array.c
gitlarryf: Good call.  Removed expandArray(), in favor of resizeArray().

PR #436, opened 2022-11-14
Implement global function pointer cache in cnex

gitlarryf:


exec/cnex/cnex.c
```diff
@@ -224,6 +224,11 @@ int main(int argc, char* argv[])
                         "Max Framesets          : %d\n"
                         "Max Allocations        : %zu\n"
                         "Remaining Objects      : %zu\n"
+                        "Total Predefined Calls : %zu\n"
+                        "Predef Linerar lookups : %zu\n"
```
ghewgill, line 228: Spelling "Linerar"

exec/cnex/cnex.c
```diff
@@ -1371,10 +1388,19 @@ void exec_CALLP(TExecutor *self)
     _IDEC_glbflags = 0;
     unsigned int start_ip = self->ip;
     self->ip++;
+     g_executor->diagnostics.total_predef_calls++;
```
ghewgill, line 1391: This line does not appear to be indented correctly

exec/cnex/global.c
```diff
@@ -481,19 +481,19 @@ Cell *global_getVariable(const char *pszVar)
...
+void (*global_lookupFunction(const char *pszFunc, uint32_t *index))(struct tagTExecutor *exec)
 {
     uint32_t i;
 
     i = 0;
     while (gfuncDispatch[i].name) {
         if (strcmp(pszFunc, gfuncDispatch[i].name) == 0) {
-            (*gfuncDispatch[i].func)(exec);
-            return;
+            *index = i;
```
ghewgill, line 491: I'm unclear here on why you need to return both the function pointer *and* its index. Wouldn't it be enough to just return one of them?

exec/cnex/module.c
```diff
@@ -114,6 +123,15 @@ TModule *module_loadModule(TExecutor *exec, const char *name, unsigned int modul
     path_readModule(r);
     bytecode_loadBytecode(r->bytecode, r->code, r->codelen);
 
+    // Allocate the global function pointer cache for this module
```
ghewgill, line 126: This block of code looks substantially similar to the above block for the main module. Is there a way to consolidate this code into one function?

ghewgill (changes requested):

exec/cnex/global.c
gitlarryf: Returning the index is just a diagnostic thing.  It tells me how many linear lookups we had to do to execute the code.  It's not _entirely_ necessary, but I'd like to keep it there, at least for a while.  Probably pull it out at some point.

exec/cnex/cnex.c
gitlarryf: Fixed.

exec/cnex/cnex.c
gitlarryf: Fixed.

exec/cnex/module.c
gitlarryf: Yea.  Removed it from module entirely, and moved it to cnex.c.

exec/cnex/global.c
gitlarryf: My only other option is to continue to pass a pointer to the `TExecutor` into this function, so I don't have to return the index.  Is that the better option?

PR #444, opened 2024-01-06
Fix cnex to deal with certain Linux variations on time/clock handling.

gitlarryf:
Fix cnex to use gettimeofday() for those Linux implementations that don't support clock_gettime().

### DIFF
--- a/exec/cnex/lib/time_linux.c
+++ b/exec/cnex/lib/time_linux.c
@@ -1,6 +1,10 @@
 #define _POSIX_C_SOURCE 200809L
 #include <stdlib.h>
+#ifdef HAVE_CLOCK_GETTIME
 #include <time.h>
+#else
+#include <sys/time.h>
+#endif
 
 #include "cell.h"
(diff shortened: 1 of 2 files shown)


exec/cnex/lib/time_linux.c
```diff
@@ -1,6 +1,10 @@
 #define _POSIX_C_SOURCE 200809L
 #include <stdlib.h>
+#ifdef HAVE_CLOCK_GETTIME
```
ghewgill, line 3: This doesn't work. On a system that does have clock_gettime (and cmake detects so), this compiler define still isn't set. Cmake variables are not the same thing as compiler defines.

ghewgill (changes requested):

exec/cnex/lib/time_linux.c
gitlarryf: Added define in CMakeLists.txt

ghewgill (approved):

PR #445, opened 2024-01-22
Set Neon to return meaningful socket errors on Windows during socket.recv() calls.

gitlarryf:
Fix net.neon to return meaningful errors on recv()


src/socketx.h
```diff
@@ -3,6 +3,7 @@
 #include <winsock.h>
 typedef int socklen_t;
 #pragma warning(disable: 4127) // incompatible with FD_SET()
+#define socketerror  WSAGetLastError()
```
ghewgill, line 6: Use `#define socketerror()` so the use of this doesn't look like a variable.

src/socketx.h
```diff
@@ -15,5 +16,6 @@ typedef int socklen_t;
 typedef int SOCKET;
 const int INVALID_SOCKET = -1;
 inline void closesocket(int x) { close(x); }
+#define socketerror   errno
```
ghewgill, line 19: Use `#define socketerror()` so the use of this doesn't look like a variable.

lib/net.cpp
```diff
@@ -15,6 +15,26 @@
...
+#ifdef _WIN32
+    FormatMessage(FORMAT_MESSAGE_FROM_SYSTEM | FORMAT_MESSAGE_IGNORE_INSERTS,
+                    NULL,
+                    err,
+                    MAKELANGID(LANG_NEUTRAL, SUBLANG_DEFAULT),
+                    msgbuf,
+                    sizeof(msgbuf),
+                    NULL);
+#else
+    snprintf(msgbuf, sizeof(msgbuf), "%s", strerror(err));
```
ghewgill, line 30: To avoid an unnecessary copy, this case can be simply `return strerror(err)`. You should be aware of default constructors and how it is unnecessary to explicitly call `std::string(...)` for character buffers.

lib/net.cpp
```diff
@@ -15,6 +15,26 @@
 
 // https://wiki.openssl.org/index.php/SSL/TLS_Client
 
+static std::string socket_error(int err)
+{
+    static char msgbuf[256] = { '\0' };
+#ifdef _WIN32
+    FormatMessage(FORMAT_MESSAGE_FROM_SYSTEM | FORMAT_MESSAGE_IGNORE_INSERTS,
```
ghewgill, line 22: The return value of FormatMessage is not checked here, and it should be.

lib/net.cpp
```diff
@@ -15,6 +15,26 @@
...
+                    NULL,
+                    err,
+                    MAKELANGID(LANG_NEUTRAL, SUBLANG_DEFAULT),
+                    msgbuf,
+                    sizeof(msgbuf),
+                    NULL);
+#else
+    snprintf(msgbuf, sizeof(msgbuf), "%s", strerror(err));
+#endif
+    if (msgbuf[0] == '\0') {
```
ghewgill, line 32: This seems like a redundant test, as I can see no case where the error message function succeeds but still returns an empty message. Instead, check for success of FormatMessage. (Note that strerror() has no error return value and is documented to always return a valid string.)

ghewgill (changes requested):

src/socketx.h
gitlarryf: Fixed.

src/socketx.h
gitlarryf: Fixed.

lib/net.cpp
gitlarryf: Fixed.

lib/net.cpp
gitlarryf: Fixed.

lib/net.cpp
gitlarryf: Removed unnecessary check, and simplified the logic.

src/socketx.h
ghewgill: This has not been fixed.

src/socketx.h
ghewgill: Also this has not been fixed.

ghewgill (changes requested):

lib/net.cpp
```diff
@@ -15,6 +15,25 @@
 
 // https://wiki.openssl.org/index.php/SSL/TLS_Client
 
+static std::string socket_error(int err)
```
ghewgill, line 18: You seem to have something named `socketerror` and something else named `socket_error`. Please find names that are more different.

ghewgill (changes requested):

lib/net.cpp
gitlarryf: I changed this to `error_to_string()`, and the macro to `GetSocketErro()`.

lib/net.cpp
```diff
@@ -15,6 +15,25 @@
 
 // https://wiki.openssl.org/index.php/SSL/TLS_Client
 
+static std::string error_to_string(int err)
+{
+#ifdef _WIN32
+    static char msgbuf[256] = { '\0' };
```
ghewgill, line 21: Why is this declared static? It doesn't need to be static.

src/socketx.h
```diff
@@ -15,5 +16,6 @@ typedef int socklen_t;
 typedef int SOCKET;
 const int INVALID_SOCKET = -1;
 inline void closesocket(int x) { close(x); }
+#define GetSocketError()   errno
```
ghewgill, line 19: This naming convention is inconsistent. Nothing else in the socket library uses CamelCase. Choose something consistent.

lib/net.cpp
```diff
@@ -15,6 +15,25 @@
 
 // https://wiki.openssl.org/index.php/SSL/TLS_Client
 
+static std::string error_to_string(int err)
```
ghewgill, line 18: This name is too generic - what kind of error? This is specifically related to sockets, so socket should be mentioned in the name somewhere.

ghewgill (changes requested):

lib/net.cpp
gitlarryf: Removed static.

src/socketx.h
gitlarryf: Changed name to socket_error()

lib/net.cpp
gitlarryf: Changed name to socket_error_message()

PR #446, opened 2024-02-15
Add meaningful Win32 socket error code to cnex

gitlarryf:


exec/cnex/lib/net.c
```diff
@@ -58,6 +58,25 @@ static SOCKET *check_socket(TExecutor *exec, Object *o)
     return o->ptr;
 }
 
+static char *socket_error_message(int err)
```
ghewgill, line 61: This should return `const char` instead of `char`

ghewgill (changes requested):

exec/cnex/lib/net.c
gitlarryf: That's what was intended.  I missed that.  It's been fixed and pushed.

PR #452, opened 2024-04-11
Add math.min function support to csnex

gitlarryf:


exec/csnex/lib/math.cs
```diff
@@ -59,6 +59,17 @@ public void log()
             Exec.stack.Push(Cell.CreateNumberCell(Number.FromDouble(r)));
         }
 
+        public void min()
```
ghewgill, line 62: If you're going to add min, then add max too.

ghewgill (changes requested):

